PR #32, opened 2023-05-16
Unify codes

heliocastro:
Unify two initial codebases


heliocastro:
@KoukiHama This is a breaking change, but i need start to unify both Toshiba and Siemens code otherwise start to became too different.
Let's discuss this on the official meeting

KoukiHama (approved):

PR #38, opened 2023-05-25
feat(home-page): Added list of releases under my comp subscriptions

amritkv:
Added list of releases under my comp subscriptions along with list of components

PR #47, opened 2023-05-26
Entire home redesign

heliocastro:


heliocastro:
@amritkv @hoangnt2 Take a look on this change.

I reduced the multiple calls to a single one, still not working due the table building, but is already a template on how to do it.
Some things are still not well defined, like the typing for the resulting data, but the code is way more readable.

@amritkv The tables code are exactly the same, so think over create a HomeTable components, making the page itself smaller in terms of code.

heliocastro:
@amritkv @hoangnt2 Please take a look on this new update.
The current home page uses a external Table as component, the fetch was over simpliefied and the code can become componentized now.

Let's use as the base for writing the pages code, and from there we can improve

amritkv:
@heliocastro Then I will build on top of this code. For both home and project page. 
And as the template has changed now, so then I think we can close the other PR i.e. #38

amritkv:
@heliocastro Are you going to merge this branch ? If not then I have to refactor the existing code on the basis of the changes suggested in this branch.

heliocastro:
@amritkv Let me finish the base page then we merge it and you go after that.
Will be in one hour or so

amritkv:
@heliocastro Ok. Waiting for that.

KoukiHama (dismissed):

ag4ums (approved):
code looks good

PR #49, opened 2023-05-30
Prepare initial table common layout

heliocastro:
Tables now based on gridjs has own layout. Is based on customization of mermaid original gridjs layout.

All components based on GrodJS and GridJS itself will rely on this base.


KoukiHama (approved):

heliocastro:
> @heliocastro, there some errors with the imports (header and footer) in main layout. Could you fix them in this PR? image

Fixed. Just noticed. Weird Mac directory miscreation

PR #57, opened 2023-06-19
Release/move components of component index

hoangnt2:
### Move source:
- Move **SideBar** to **/components/sw360**
- Move related components of components page from **src/components/component** to **src/app/[locale]/components/components** 

### Re-define **AdvanceSearch** fields format:
Now search fields will be contained in an array of object with format:
```
    [
            {
                fieldName: 'Component Name',
                value: '',
                paramName: 'name'
            },
            {
                fieldName: 'Group',
                value: [
                    {
                        key: 'None',
                        text: 'None'
                    }
                ],
                paramName: 'group'
            }
    ]
```

**fieldName**: field's label,
**paramName**: name of params on URL,
**value**: can be **string** or an **array of options**
 + **value** is **string**: it will be **text box.**
 + **value** is an **array** of options: it will be **select box**. The **options** will have below format, with **key** is the **value of option** and text is **displayed text:**
 
```
        {
            key: 'CODE_SNIPPET',
            text: 'Code Snippet'
        }
```
The option will look like this
image

Apply handle event on **AdvanceSearch**

Replace **ComponentAdvanceSearch** on Component Index page by **AdvanceSearch**


heliocastro (approved):
Now we're talking :-)

PR #58, opened 2023-06-22
fix(home-page): Added search icon and fixed empty data in my subscrip…

amritkv:
1. Added search icon in header component
2. Fixed the empty data issue in my subscriptions (components & releases) column


heliocastro (approved):
Good to go

PR #60, opened 2023-06-23
Handle some buttons's event and add missing parts of vulnerabilities tab on Component detail page

hoangnt2:
- Handle delete release on **Release Overviews** tab

Delete Release

- Handle download attachment and attachment bundle on **Attachments** tab

Download Attachment

- Add missing parts and handle change state  on tab **Vulnerabilities**:
      1. Vulnerability matching statistics
      2. Tool Tip when hover mouse on verification column
      3. Now we can change state of vulnerabilities

Vulnerabilities


heliocastro (approved):
Let's go this way for now, but in the future we move all components to the sw360 namespace

PR #61, opened 2023-06-23
fix(home-page): Added default values for every table, subscriptions a…

amritkv:
Added default values for all the tables, my subscriptions, recent releases and recent components at home page

### DIFF
--- a/src/app/[locale]/home/components/MyComponentsWidget.tsx
+++ b/src/app/[locale]/home/components/MyComponentsWidget.tsx
@@ -1,4 +1,5 @@
 // Copyright (c) Helio Chissini de Castro, 2023. Part of the SW360 Frontend Project.
+// Copyright (C) Siemens AG, 2023. Part of the SW360 Frontend Project.
 
 // This program and the accompanying materials are made
 // available under the terms of the Eclipse Public License 2.0
@@ -16,6 +17,7 @@ import HomeTableHeader from './HomeTableHeader'
 
 import { Table } from '@/components/sw360'
 import { sw360FetchData } from '@/utils/sw360fetchdata'
+import defaultMyComponents from '../../../../../defaultValues/defaultValuesHome-MyComponents.json'
 
 function MyComponentsWidget() {
     const [data, setData] = useState([])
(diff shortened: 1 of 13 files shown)


src/app/[locale]/home/components/MyComponentsWidget.tsx
```diff
@@ -24,7 +26,14 @@ function MyComponentsWidget() {
     useEffect(() => {
         const fetchData = async () => {
             const data = await sw360FetchData('/components/mycomponents', 'components')
-            setData(data.map((item: { name: string, description: string }) => [item.name, item.description]))
+            if (data.length > 0)
```
Comment (hoangnt2) [inline src/app/[locale]/home/components/MyComponentsWidget.tsx:29]: You can write this code snippets like this. Same with other files
```
if (data.length > 0) {
data = defaultMyComponents
}
setData(data.map((item: { name: string, description: string }) => [item.name, item.description]))
```

Comment (heliocastro) [reply src/app/[locale]/home/components/MyComponentsWidget.tsx:29]: Should we have defaultValues ? Should not be the string for empty ?
Example:
https://gridjs.io/docs/config/className -> Look for notFound

And gridjs code already has a default message for empty:
```
      {status === Status.Rendered && data && data.length === 0 && (
        <MessageRow
          message={_('noRecordsFound')}
          colSpan={headerLength()}
          className={classJoin(
            className('notfound'),
            config.className.notfound,
          )}
        />
      )}
```

Comment (amritkv) [reply src/app/[locale]/home/components/MyComponentsWidget.tsx:29]: This also looks fine. But I was thinking of showing table specific messages

For example : 

For `My Projects` table : We can show, `No projects found` , or 
For `My Components` table : We can show, `No components found` 

What you people think about it ?

Comment (hoangnt2) [reply src/app/[locale]/home/components/MyComponentsWidget.tsx:29]: @amritkv. In my opinion, no record found message is enough

ag4ums:
current sw360 the message is 
image

heliocastro:
Let me check one simple thing, maybe we can solve this with a single line instead this overblown code.

heliocastro:
This pul request is now invalidated by #90

PR #65, opened 2023-07-04
Move components and layout

heliocastro:
- Hide navbar when user not logged in
- Move some components to SW360 namespace
- Change keys in messages. The key should not have dots. Need to verify a proper naming for the keys, even not be optimal
- Update dependency versions


messages/en.json
```diff
@@ -122,7 +122,7 @@
         "License Shortname": "License Shortname",
         "License Type": "License Type",
         "Logout": "Sign Out",
-        "Learn more about component types.": "Learn more about component types.",
+        "learn_more_components": "learn_more_components",
```
hoangnt2, line 125: Keep translated value "Learn more about component types."

hoangnt2:
@heliocastro, I have one comment only. Everything look good

messages/en.json
heliocastro: We need come out with ids.
The value itself is the key, not the resulting one, so TS is complaining about having keys with dot.

So in the end, we have to standardize some ID for the messages, otherwise we will end up with this mess.

I know somehow is harder, but we shou;d come with something like MSG_123 AND separate the messages in their own namespaces, like global, components.

hoangnt2: @heliocastro, I understood. The key cannot contain dot. I mean change value to "Learn more about component types."

```
"learn_more_components": "Learn more about component types.",
```

hoangnt2 (approved):

PR #69, opened 2023-07-19
Fix Gridjs error and warning logs, add missing page size selector for component and release page

hoangnt2:
### Fix the following Gridjs's error and warning logs
 1. [Grid.js] [ERROR] Duplicate plugin ID: pagination
 2. [Grid.js] [ERROR] Duplicate plugin ID: search
 3. [Grid.js] [ERROR]: The container element [object HTMLDivElement] is not empty. Make sure the container is empty and call render() again
 4. [Warning] ReactDOM.render is no longer supported in React 18
### Add missing page size selector for component and release page


heliocastro (approved):
This is Perfect.
Many thanks @hoangnt2 to tkae care of it.

PR #70, opened 2023-07-21
feat(fossology): Handle fossology reload report

hoangnt2:
## Handle fossology reload report

test

 This function will work when https://github.com/eclipse-sw360/sw360/pull/2047 was merged


heliocastro (approved):
Good to go

PR #73, opened 2023-08-01
Handle export spreadsheet component, fix styles and data in component details page

hoangnt2:
## Handle export spreadsheet component, fix styles and data in component details page
**Changes:**
- Component Home page:
   + Export spreadsheet component with releases and without releases
- Component details page:
   + Fix styles
   + Fix moderators and vendors data


heliocastro (approved):
GTG

PR #76, opened 2023-08-07
feat(component): Edit Component page 

tuannn2:
Changes:
   Feature edit Component
         - Tab Sumary
         - Tab Release
         - Tab Attachments


tuannn2:
Thanks for your comment, @heliocastro.
I have split this commit in small atomic commits.
Please check it.

src/app/[locale]/components/edit/[id]/components/EditComponent.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    const handleDeleteComponent = () => {
+        setDeleteDialogOpen(true)
+    }
+
+    const headerButtons = {
+        'Update Component': { link: '/components/edit/' + componentId, type: 'primary', onClick: submit },
+        'Delete Component': {
+            link: '/components/edit/' + componentId,
+            type: 'primary',
```
Comment (hoangnt2) [inline src/app/[locale]/components/edit/[id]/components/EditComponent.tsx:143]: type danger

src/app/[locale]/components/edit/[id]/components/Releases.tsx
```diff
@@ -0,0 +1,104 @@
...
+            ) {
+                const data = releaseLinks['_embedded']['sw360:releaseLinks'].map((item: ReleaseLink) => [
+                    item.name,
+                    [item.id, item.version],
+                ])
+                setData(data)
+            }
+        })
+    }, [])
+
```
Comment (heliocastro) [inline src/app/[locale]/components/edit/[id]/components/Releases.tsx:65]: @hoangnt2 @amritkv Can we get in a common way to fetch data ? Right now i'm using @hoangnt2 APIUtils on my license code, but there are at least three different ways of fetching data.
We should set for one otherwise will become way too hard to debug

hoangnt2:
@tuannn2 please verify your commits

src/app/[locale]/layout.tsx
```diff
@@ -12,6 +12,7 @@ import 'bootstrap/dist/css/bootstrap.min.css'
 import '@/styles/globals.css'
 import '@/styles/auth.css'
 import '@/styles/gridjs/sw360.css'
+import 'react-toastify/dist/ReactToastify.css'
```
Comment (heliocastro) [inline src/app/[locale]/layout.tsx:15]: I like Toastify, but for the sake of consistency, can you look if we have something in react-bootstrap ?

src/components/Attachments/SelectAttachment/SelectAttachment.tsx
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        const url = SW360_API_URL + '/resource/api/attachments'
+        fetch(url, {
+            method: 'POST',
+            body: formData,
+            headers: {
+                Authorization: `Bearer ${session.user.access_token}`,
+            },
+        })
```
heliocastro, line 77: @hoangnt2 @amritkv See, another place passing direct Bearer instead of use the global Session object

src/app/[locale]/components/edit/[id]/components/EditComponent.tsx
```diff
@@ -0,0 +1,195 @@
...
+import CommonTabIds from '@/object-types/enums/CommonTabsIds'
+import ApiUtils from '@/utils/api/api.util'
+import { Session } from '@/object-types/Session'
+import HttpStatus from '@/object-types/enums/HttpStatus'
+import { signOut } from 'next-auth/react'
+import { notFound, useRouter } from 'next/navigation'
+import { useTranslations } from 'next-intl'
+import { COMMON_NAMESPACE } from '@/object-types/Constants'
+import { SideBar, PageButtonHeader } from '@/components/sw360'
+import { toast, TypeOptions, ToastContainer } from 'react-toastify'
```
Comment (heliocastro) [inline src/app/[locale]/components/edit/[id]/components/EditComponent.tsx:23]: I like Toastify, but for the sake of consistency, can you look if we have something in react-bootstrap ?

heliocastro (changes requested):
Mentioning @hoangnt2 and @amritkv 
We do need a global way to fetch code, can you too start a proposal on discussions ?

@tuannn2 Code is generaly ok, but look for my remarks. The fetch code we deal later.

hoangnt2:
> Mentioning @hoangnt2 and @amritkv We do need a global way to fetch code, can you too start a proposal on discussions ?
> 
> @tuannn2 Code is generaly ok, but look for my remarks. The fetch code we deal later.

My APIUtils works with any request methods (GET, POST, PATCH...). We can simply create a service and use APIUtils to call a GET request in it.

src/app/[locale]/components/edit/[id]/components/EditComponent.tsx
Comment (tuannn2) [reply src/app/[locale]/components/edit/[id]/components/EditComponent.tsx:23]: thanks, I will look into it and update it along with Release Function. After the PR is merged.

amritkv:
> 

@heliocastro @hoangnt2 Here we need to decide whether we want send the entire fetched data to the called function or only the required part of the fetched data.

hoangnt2:
> 
@heliocastro, @amritkv I think fetch function should return data and response status instead of just data, sometimes we need to use state to handle some special cases

tuannn2:
Thanks for your comment, @hoangnt2,
I updated and verified commits.

hoangnt2:
@tuannn2 Please resolve error and warning messages
image

ComponentAddSummary, ComponentEditSummary aren't global components. So please move them from **/components/sw360** to **/components**

tuannn2:
Thanks @hoangnt2,
Done.

hoangnt2 (approved):

heliocastro (approved):
GTG

PR #83, opened 2023-08-22
feat(project): Added external ids integration while creating project

amritkv:
Feature : Added the integration of external ids while creating a project


hoangnt2:
@amritkv The code looks good.
But when we click Create Project button with blank empty for required fields, they should show the message.
And when I create project successfully, the alert function doesn't seem to work
So please fix them in the next PRs

hoangnt2 (approved):

PR #85, opened 2023-08-23
feat(project_detail_summary): Project detail summary page integration

deo002:
Project Detail Summary Page Integration


src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
```diff
@@ -0,0 +1,269 @@
...
+            vendorShortName: "",
+            vendorUrl: ""
+        }
+    )
+
+    useEffect(() => {
+        const controller = new AbortController()
+        const signal = controller.signal
+
+        ;(async () => {
```
Comment (hoangnt2) [inline src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:77]: To much requests to fetch project summary data. Need to re-define API endpoint to fetch the necessary data simultaneously

src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
```diff
@@ -0,0 +1,269 @@
...
+        })()
+
+        return () => controller.abort()
+    }, [])
+
+    return (
+        <>
+            <div className="ms-5 mt-2">
+                <Tab.Container defaultActiveKey="summary">
+                    <Row>
```
Comment (hoangnt2) [inline src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:192]: I created a global component to define Sidebar menu you can refer to component detail page

src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+import { SummaryDataType } from '@/object-types/SummaryDataType'
+import { notFound } from 'next/navigation'
+
+export default function ViewProjects({ session, projectId }: { session: Session, projectId: string }) {
+
+    const t = useTranslations(COMMON_NAMESPACE)
+    const [loading, setLoading] = useState(true)
+    const [summaryData, setSummaryData] = useState<SummaryDataType>(
+        {
```
Comment (hoangnt2) [inline src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:34]: You can set init value for summaryData is **undefined** then detail component is only shown when the summary data is not **undefined**

src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:77]: Yes, too many requests. The API endpoint will be changed to reduce the number of requests.

src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:34]: Yep! Sounds good!

src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:192]: I feel that the React-Bootstrap Tab component is more generic and neat to use. I was wondering if we could we shift to that? What do you suggest?

Comment (heliocastro) [reply src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx:192]: I agree with @deo002 
@hoangnt2 The global SideBar component was still not optimal and is not generic at all, making really har to portable usage.
I looked around to see how both bootstrap react Tab and NextUI Tab, and they are implemented in a neat separation way, so i do suggest we go with Bootstrap version for now

heliocastro (approved):
Let's use Bootstrap Tab

heliocastro:
@deo002 Please resolve the conflicts

deo002:
Opened an issue for the API changes: eclipse/sw360#2108

deo002:
Can this be reviewd and merged so that the UI work goes on? I'll incorporate the API changes as and when they happen. @heliocastro @hoangnt2

heliocastro:
@deo002 Next time don't forget to unmark the label conflict, otherwise we will never know that is fixed

heliocastro (approved):
Looks good.

@deo002 For the future, verify if we can skip use <Row> <Col> and keep using only Tab.* from bootstrap

heliocastro:
@deo002 I can't merge, you need sign your commits

deo002:
Signed it!

> @deo002 I can't merge, you need sign your commits

PR #88, opened 2023-08-24
 feat(release): Add tab Commercial Details 

tuannn2:


heliocastro (approved):
Looks good, but please, do a post commit moving the CommercialDetails from global components to inner Components tab.
Or this is suppose to be global ?

tuannn2:
Thank you for your comment, @heliocastro  
Commercial Details used only in Release. 
I will move to inner Release tab.

PR #92, opened 2023-08-25
Create tree table derived component

heliocastro:
This component provides capacity to current table to extend internal rows as a tree. Is most used as the components and packages subtree inside SW360.

Special thanks to @hoangnt2 that finished the initial proof of concept.


hoangnt2:
This function can work properly with small trees but I am concerned about performance issues when the data of tree is too large. @heliocastro @amritkv  Do you have any ideas or solutions?

heliocastro:
> This function can work properly with small trees but I am concerned about performance issues when the data of tree is too large. @heliocastro @amritkv Do you have any ideas or solutions?

The data could be preprocessed on server side, right now we have this solution, but when we got alpha out of the door, we will start to make the "real work™", meaning, is clear that what we are doing right now is not the final code, because we don't know yet what we want in final code. 
So we will need write a document with the right questions for the inevitable refactor THEN we can say let's go to beta  and production.

PR #93, opened 2023-08-25
fix(project_add_summary): Fixed on hover messages and required field warnings

amritkv:
1. Fixed the on hover message structure.
2. Fixed required fields' missing value warning which was not appearing in prev, commits.

Note: Please do not merge it now. Need to push a little more change


src/app/[locale]/projects/add/page.tsx
```diff
@@ -15,7 +15,7 @@ import { useTranslations } from 'next-intl'
 import { COMMON_NAMESPACE } from '@/object-types/Constants'
 import { useRouter } from 'next/navigation'
 import HttpStatus from '@/object-types/enums/HttpStatus'
-import { TypeOptions, toast } from 'react-toastify'
+import { ToastContainer, TypeOptions, toast } from 'react-toastify'
```
Comment (heliocastro) [inline src/app/[locale]/projects/add/page.tsx:18]: Again. I already mentioned that:
Or we assume use along everywhere toastify or we use the ones already available on react-bootstrap.
@hoangnt2 @amritkv We need reach a proper decision.

For the sake of maintenance and consistency, i do not want to have two things doing the same thing on the code.

Comment (heliocastro) [reply src/app/[locale]/projects/add/page.tsx:18]: This https://react-bootstrap.github.io/docs/components/toasts or Toastify

heliocastro (approved):
Approved now, but with a caveat, we need to be consistent.
After the decision, we should write it as set on stone for v1

src/app/[locale]/projects/add/page.tsx
Comment (amritkv) [reply src/app/[locale]/projects/add/page.tsx:18]: If it is readily available out of react-bootstrap then we can stick to it instead of external library i.e. react-toastify.
What you think @heliocastro and @hoangnt2 ?

Comment (hoangnt2) [reply src/app/[locale]/projects/add/page.tsx:18]: @heliocastro @amritkv I agree with this, let's use react-bootstrap's toast

heliocastro (changes requested):
@amritkv Please remove toastify and use bootstrap react toast

heliocastro (approved):
LGTM

PR #115, opened 2023-09-12
feat(link_projects): Integration of functionality to link projects

deo002:
Integration of functionality to link projects


src/object-types/enums/HttpStatus.ts
```diff
@@ -19,6 +19,7 @@ enum HttpStatus {
     NOT_FOUND = 404,
     CONFLICT = 409,
     INTERNAL_SERVER_ERROR = 500,
+    FORBIDDEN = 403
```
hoangnt2, line 22: keep this in order

hoangnt2 (changes requested):

src/object-types/enums/HttpStatus.ts
deo002: Done

hoangnt2 (approved):
Changes look good

PR #116, opened 2023-09-12
Fix styling

hoangnt2:
1. Fix styling bug of inputs
2. Replace exact match tool tip by ShowInfoOnHover component
3. Add scroll bar to LinkReleaseToProjectModal when the content is too large

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -118,7 +118,7 @@ footer {
     padding: 10px;
 }
 
-.table thead th{
+.table thead th {
     background-color: var(--sw360-primary-background-color);
     color: white;
     font-size: bold;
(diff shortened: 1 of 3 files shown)


src/styles/globals.css
```diff
@@ -393,24 +393,8 @@ th {
...
-    font-size: 0.875rem;
-    display: block;
-    transition: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
-    font-weight: 400;
-    height: 1.5rem;
-    padding-bottom: 0.25rem;
-    padding-left: 0.75rem;
-    padding-right: 0.75rem;
-    padding-top: 0.25rem;
-}
```
heliocastro, line 411: Why remove entire form.control ? This is to be exactly like liferay styling, incluDing the orange fading in.

heliocastro (commented):
Some commentsLook for

src/styles/globals.css
hoangnt2: @heliocastro This snippet made the input look smaller. I am not sure why we need this. I think bootstrap's default  is enough

With form-control
image

Without form-control
image

heliocastro: Ok, got the point

heliocastro (approved):
LGTM

PR #124, opened 2023-09-26
Fix eslint errors for components in component details page

hoangnt2:
Fix eslint errors:
- Component table
- Advance Search
- Delete Component Modal
- Import SBOM modal


heliocastro (approved):
I'm really pleased.
Many thanks for this effort.

PR #127, opened 2023-09-27
feat(eslint): Fix eslint errors for Edit Release page

tuannn2:
### Fix eslint errors:

- Function common uitls
-  Get release for edit release


heliocastro (approved):
Very good

PR #133, opened 2023-10-03
refactor(project): Start planned major refactor towards alpha release

heliocastro:
The first entry of the major refactors is undergoing to reduce code duplication and multiple fetch entries.
The end goal is to have a single fetch process and take away all the boilerplate that has been duplicated over and over in the pages.


hoangnt2:
The changes look good to me

src/components/Attachments/EditAttachments.tsx
```diff
@@ -101,7 +100,7 @@ const EditAttachments = ({
                 }
             }
         })
-    }, [documentId, documentType, fetchData])
+    }, [documentId, documentType, fetchData, componentPayload, setComponentPayload])
```
hoangnt2, line 103: **componentPayload* and **setComponentPayload** should be remove from useEffect's dependencies. They can cause some logic errors

hoangnt2 (dismissed):

src/components/Attachments/EditAttachments.tsx
hoangnt2: @heliocastro I know you're trying to fix the warnings, but useEffect's dependencies should not be changed.

src/components/sw360/index.ts
```diff
@@ -17,6 +17,7 @@ import QuickFilter from './QuickFilter/QuickFilter'
 import SearchUsersModal from './SearchUsersModal//SearchUsersModal'
 import SearchVendorsModal from './SearchVendorsModal/SearchVendorsModal'
 import SideBar from './SideBar/SideBar'
+import ToastMessage from './ToastContainer/Toast'
```
hoangnt2, line 20: Toast is a client component so you need to add 'use client' on top of this file or Toast.tsx. It will solve the error

heliocastro: Toast.tsx it is.

KoukiHama (approved):

PR #135, opened 2023-10-05
feat(search_page): Search page integration

deo002:


deo002:
Entire document option is not supported by API yet. Will add it when the API changes are made.

heliocastro (approved):
LGTM

heliocastro:
@deo002 Next time it can be one pull request, but you can split it into several commits.
i.e. Messages could be in "feat(messages): Some text" and so on...

PR #144, opened 2023-10-10
feat(add_project): Incorporated link project feature while creating a…

amritkv:
WIP : Incorporated link project feature while creating a project


src/app/[locale]/projects/add/components/ProjectAddSummary.tsx
```diff
@@ -20,7 +20,7 @@ import { ToastContainer } from 'react-bootstrap'
 import ApiUtils from '@/utils/api/api.util'
 import ProjectPayload from '@/object-types/CreateProjectPayload'
 import { useState } from 'react'
-import { Session } from '@/object-types'
```
Comment (heliocastro) [inline src/app/[locale]/projects/add/components/ProjectAddSummary.tsx:23]: Nope nope nope, now is { Session }

src/app/[locale]/projects/add/page.tsx
```diff
@@ -9,7 +9,7 @@
 
 import { authOptions } from '@/app/api/auth/[...nextauth]/route'
 import { getServerSession } from 'next-auth/next'
-import { Session } from '@/object-types'
```
Comment (heliocastro) [inline src/app/[locale]/projects/add/page.tsx:12]: Nope nope nope

src/components/ProjectAddSummary/LinkedReleasesAndProjects.tsx
```diff
@@ -11,13 +11,17 @@
 
 import LinkedReleases from './component/LinkedReleasesAndProjects/LinkedReleases'
 import LinkedProjects from './component/LinkedReleasesAndProjects/LinkedProjects'
-// import { Session } from '@/object-types/Session'
+import Session from '@/object-types/Session'
```
heliocastro, line 14: Nope nope nope

heliocastro (changes requested):

src/components/ProjectAddSummary/LinkedReleasesAndProjects.tsx
amritkv: @heliocastro Sorry, I didn't get your point. As the `Session.ts` has default export so the import syntax should be like this right ?
Or I'm in other direction ?

heliocastro: Was, but now since you followed the refactor process, everything that is global is moving go a declared "unpack" folder,
look on src/object-types/index.ts

And this is defined here in tsconfig.json:
```json
        "paths": {
            "next-sw360": ["./src/components/sw360"],
            "@/app/*": ["./src/app/*"],
            "@/assets/*": ["./src/assets/*"],
            "@/components/*": ["./src/components/*"],
            "@/messages": ["./messages"],
            "@/object-types": ["./src/object-types"],
            "@/object-types/*": ["./src/object-types/*"],
            "@/services/*": ["./src/services/*"],
            "@/styles/*": ["./src/styles/*"],
            "@/utils/*": ["./src/utils/*"],
            "@/utils": ["./src/utils"]
        },
```

So, your changes work now, but will be deprecated in the future for globals.
Now works the two ways, direct like you changed or unpacked:

```json
            "@/object-types": ["./src/object-types"],
            "@/object-types/*": ["./src/object-types/*"],
```

With the time, we will move non global types to their page area, to avoid this huge amount of info there

amritkv: @heliocastro Please review once.

heliocastro (approved):
Now is good to go. You need to remove the TAG WIP

heliocastro (approved):

PR #154, opened 2023-10-18
fix(Component): Gridjs pagination server side for component page

hoangnt2:


heliocastro (approved):
Few minor changes needed, but this will be sorted out in the refactoring process ongoing.
Really great work @hoangnt2

PR #158, opened 2023-10-24
fix(session): Fix session and error provider for component, release 

tuannn2:
- Fix session not found when call api
- Fix error provider for component, release


heliocastro (approved):
Good to go

PR #159, opened 2023-10-25
refactor(projects_table): Add server side pagination to projects table

deo002:


heliocastro (approved):
Good to go.
Maybe we should consider reduce the boilerplate of the code and check the session direct on table component, so no need to add headers

deo002:
> Good to go. Maybe we should consider reduce the boilerplate of the code and check the session direct on table component, so no need to add headers

Yes! That will be good but we would have to find a way for it...for now, we cannot use hooks(useSession in this case) in a class component(Table).

PR #162, opened 2023-10-29
Add admin page

deo002:


heliocastro (approved):
@hoangnt2 ( to keep notified )
I'm approving this commit since we're pre-alpha stages, but somehow the whole navbar became overcomplicated, and we need to decompose and make it proper.
But not now, let have all components around, and we start to refine later.
@deo002 Please open a discussion about navbar improvements and then associate an issue to fix it in the future.

deo002:
> @hoangnt2 ( to keep notified ) I'm approving this commit since we're pre-alpha stages, but somehow the whole navbar became overcomplicated, and we need to decompose and make it proper. But not now, let have all components around, and we start to refine later. @deo002 Please open a discussion about navbar improvements and then associate an issue to fix it in the future.

Will do it!

PR #172, opened 2023-11-17
Check session

deo002:


deo002:
I'll add too other pages as well once this is approved

hoangnt2:
@deo002  I'm wondering where we can put it. It is a component but HOC is quite different in meaning.

deo002:
> @deo002 I'm wondering where we can put it. It is a component but HOC is quite different in meaning.

Hmm...how about a folder named HOCs inside of components/sw360 and import it like this: import 'withAuth' from '@/components/sw360/HOCs'.

Also, I was considering renaming withAuth to loadSession.

What do you think?

hoangnt2:
@deo002 I think **hocs** should be on same level with the **components** folder.

image

Here is example from ChatGPT

and **withAuth** should be kept (I had taken a look on some HOC example and they use that name format **with + sth**)

heliocastro:
I'm still not 100% convinced though.
Is higher order component the correct solution and / or the only ?

hoangnt2:
@heliocastro

The routes are protected by middleware.ts (as we configured in the project).

The problem with useSession is when you use useSession it will fetch the session from api/auth/session and it will take some time to wait so we need a piece of code to check if the session is fetched or not. I'm using HOC to reduce code duplication (check session code), not to protect routes.

deo002:
@heliocastro The hoc is not for protecting routes, it's sole purpose is to wait(show spinner) until the session info is loaded and then provide it to the components that use it. 

If we use session before it has loaded, we get this error because session.user.token is undefined until then: 
Screenshot from 2023-11-06 11-13-05

We can discuss alternative approaches or how to improve this approach this week.

deo002:
> @deo002 I think **hocs** should be on same level with the **components** folder.
> 
> image
> 
> Here is example from ChatGPT
> 
> and **withAuth** should be kept (I had taken a look on some HOC example and they use that name format **with + sth**)
> 
> @heliocastro could you give me and @deo002 some advices?

Sounds good!

heliocastro:
> 
> We can discuss alternative approaches or how to improve this approach this week.

useSession has a intermediate state ( loading ), to do exactly the timing difference. We do need discuss this one, before we overcomplicate all.

deo002:
> > We can discuss alternative approaches or how to improve this approach this week.
> 
> useSession has a intermediate state ( loading ), to do exactly the timing difference. We do need discuss this one, before we overcomplicate all.

Sure! Let's discuss.

heliocastro:
Closing this one as we are already using better solution

PR #179, opened 2023-11-30
Fix page styling and component error, hide locale prefix from URL

hoangnt2:
This PR does several things:
- Fix some global css not working.
- Align page content
- Hide locales prefix (en, ja, ...) from URL
- Fix bug prop's value is undefined for AddAdditionalRoles


heliocastro (approved):
Looks good to me.
Hide the language endpoints was a good idea

PR #181, opened 2023-11-30
fix(Table): Fix change page size error when server pagination is not defined

hoangnt2:
Fix change page size error (License page) when server pagination is not defined


heliocastro (approved):
Very good

PR #186, opened 2023-12-13
Admin licenses1

deo002:


deo002:
The translation files are being reformatted on comitting for some reason

heliocastro (changes requested):
Please leave empty all other languages taht is not properly translated.
Adding english in the fields will make the automatic detector of missing transalation lost.
So, i do recommend leave all  the entries empty

heliocastro:
> The translation files are being reformatted on comitting for some reason

If you setup husky properly in your system, yes, files are reformatted before be commit. Is intentional.

deo002:
> Please leave empty all other languages taht is not properly translated. Adding english in the fields will make the automatic detector of missing transalation lost. So, i do recommend leave all the entries empty

Done

PR #197, opened 2024-01-04
Second batch of layout fixes

heliocastro:


src/app/[locale]/preferences/components/NotificationSettings.tsx
```diff
@@ -64,50 +61,32 @@ const NotificationSettings = ({ notificationSetting, setNotificationSetting }: P
...
-                            {Object.values(value.setting).map((entry) => (
-                                <div className='form-group' key={entry.id}>
-                                    <div className='form-check'>
-                                        <input
-                                            id={entry.id}
+                <>
+                    <Accordion.Item eventKey={value.documentType}>
+                        <Accordion.Header>{value.documentType}</Accordion.Header>
+                        <Accordion.Body>
+                            <Form name={value.documentType}>
```
Comment (hoangnt2) [inline src/app/[locale]/preferences/components/NotificationSettings.tsx:70]: Move this Form to div. This causes some errors
image

next.config.js
```diff
@@ -11,7 +11,7 @@
 const withNextIntl = require('next-intl/plugin')()
 
 const nextConfig = {
-    reactStrictMode: true,
```
hoangnt2, line 14: I'm not sure we should get rid of react strict mode

src/app/[locale]/preferences/components/NotificationSettings.tsx
Comment (heliocastro) [reply src/app/[locale]/preferences/components/NotificationSettings.tsx:70]: Fixed, as the entries were never returned before SSR is rendered.
No need for move back to div.

next.config.js
heliocastro: Solved

src/app/[locale]/preferences/components/NotificationSettings.tsx
Comment (heliocastro) [reply src/app/[locale]/preferences/components/NotificationSettings.tsx]: Fixed

hoangnt2 (approved):
Change look good

PR #214, opened 2024-02-12
feat(project_attachment_usages): Add table for project attachments

deo002:


heliocastro (changes requested):
Please don't reformat entire en.json.
Can you redo the commit only with effective change ?

deo002:
> Please don't reformat entire en.json. Can you redo the commit only with effective change ?

It is reformatting on it's own when I do a commit. I tried a few things, but none of them are working. Can you help me with this?

heliocastro (approved):

heliocastro:
@deo002 Ok, if the linter is formatting on this way is fine. I needed deactivate prettier-quick due incompatibility of current codebase, so let's not fight over it.

PR #217, opened 2024-02-14
feat(admin_users): List all users in admin users

deo002:


deo002:
Reformatting is happening in messages/en.json while commiting. Should I exclude all the messages folder files from pre-commit?

heliocastro (approved):
LGTM

PR #218, opened 2024-02-20
fix(edit_project): Fixed ownergroup, owingaccount, country under roles

amritkv:
fix(edit_project): Fixed ownergroup, owingaccount, country under roles


heliocastro (approved):
LGTM

PR #223, opened 2024-02-23
Implement License detail page

hoangnt2:


heliocastro:
@hoangnt2  ->. ERROR one or more files does not have a valid license header

hoangnt2:
> @hoangnt2 ->. ERROR one or more files does not have a valid license header
I fixed it

PR #236, opened 2024-03-19
fix(home_page): Fixed redirection on clicking a release at home page

amritkv:
Fixed redirection on clicking a release at home page and search page


heliocastro (approved):
LGTM

PR #237, opened 2024-03-20
Fix table bugs

hoangnt2:
- Fixed bug incorrect table props are updated when the table is re-rendered
- Fixed an issue where tables using server paging were re-rendered multiple times (Project, Component, Vulnerability pages)


deo002:
Looks cool!

heliocastro (approved):
LGTM

heliocastro:
@hoangnt2 Out of curiosity, i like the idea of remove some props logic from the constructor, but is there another reason you did that ? Reuse it in extend table ?

hoangnt2:
> @hoangnt2 Out of curiosity, i like the idea of remove some props logic from the constructor, but is there another reason you did that ? Reuse it in extend table ?

@heliocastro, that logic is also used in **com ComponentDidUpdate** so I moved it out of the constructor to reuse it.

PR #239, opened 2024-03-25
Fix bug cannot edit licenses for release

hoangnt2:
##### Fixed bug:
+ Main licenses cannot be shown in release detail page
+ Function to edit main and other licenses does not working properly (this function will work when [API ](https://github.com/eclipse-sw360/sw360/pull/2380) is merged)

##### Make the licenses selection dialog component more common, remove duplicated components


heliocastro (approved):
Great code. I hope we can get rid of UserSearchParams on other places as well

hoangnt2:
> Great code. I hope we can get rid of UserSearchParams on other places as well

First, I will try to fix components and releases page bugs. Then I will find the way to solve UserSearchParams warning

PR #245, opened 2024-04-04
Add global message

hoangnt2:
#### What does this PR do?:
This PR create a new **GlobalMessage** component that will replace **ToastMessage**

newcomp

- This component allows message to show even when redirecting the page. This is impossible with **ToastMessage**
_For example: I call **show message** in **page A** then call **redirect** to **page B**, the message still **appears** in **page B**_

- We don't need to **embed** **GlobalMessage** and **create state** in the component where we want to display the message (easier to use)

#### How to use this component:
- All we need to do is use **MessageService**
`import MessageService from '@/services/message.service'`
- Then call `MessageService.[message_type](message_string, options})`
     +  **message_type** : can be **error**, **success**, **info**, **warn** (depends on your scenario)
     +  **message_string** : the string of the message that you want to show
     +  **options**: By default the message will be kept when we redirect to other page and auto close after 7 second, we can change these default options by adding **MessageOptions** to 2nd param. **MessageOptions** support 2 option
           -   **autoClose** (boolean): this allow message is auto closed after 7 second (default is **true**)
           -   **keepAfterRouteChange** (boolean): this allow the message will be kept when we redirect to other page (default is **true**)
 
 #### Example:
 1. Show success message with default options (auto close, keep showing when redirect page):
 ```
 MessageService.success('Show success message')
 ```
 
 2. Show success message with custom options (not auto close,  hide when redirect page):
  ```
  MessageService.success('Show success message', {autoClose: false, keepAfterRouteChange: false})
  ```
 #### I also replaced all **ToastMessage** in existing pages


hoangnt2:
@heliocastro, @amritkv , @deo002, please take a look on this PR

heliocastro (approved):
This is amazing.
I would say that's capture exactly the direction we want to go.

PR #246, opened 2024-04-08
Fix component user search

hoangnt2:
- Replace search users component by common component (SelectUserDialog).
- Remove unused and uncommon components
- Remove unused functions and types
- Fix styling for add/edit release/component page


heliocastro (approved):
Very good

PR #249, opened 2024-05-06
fix(licenseClearing): Fix bugs in project details license clearing table

deo002:
Fixes:
1. Undefined/null reference errors.
2. Multiple rows for same project error.
3. Forever loading when an error occurs.


heliocastro (approved):
LGTM
Next time maybe create an interface for the mapping

PR #250, opened 2024-05-06
feat(vulnerability_tracking_status): Integrate endpoint for vulnerabilityTrackingStatus

deo002:


src/app/[locale]/projects/detail/[id]/components/VulnerabilityTrackingStatus.tsx
```diff
@@ -0,0 +1,128 @@
...
+            sort: true,
+        },
+        {
+            id: '   vulnerabilityTrackingStatus.type',
+            name: t('Type'),
+            sort: true,
+        }
+    ]
+
+    const initServerPaginationConfig = (session: Session) => {
```
Comment (heliocastro) [inline src/app/[locale]/projects/detail/[id]/components/VulnerabilityTrackingStatus.tsx:58]: Do not pass session, use the hook

heliocastro (changes requested):
Just avoid pass session as reference and rest is ok

src/app/[locale]/projects/detail/[id]/components/VulnerabilityTrackingStatus.tsx
Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/VulnerabilityTrackingStatus.tsx:58]: We are using hook only but passing on session like that to tackle the multiple rerendering of tables. PR #237

heliocastro (approved):
Ok, mark for future verify all session usage in system

PR #253, opened 2024-05-07
feat(moderation_request): Added feature to unassign moderator

amritkv:
Added feature to unassign moderator from moderation request
Minor Fix: Changed document name column (open and closed moderation request table) width to auto to accommodate long names.


heliocastro (approved):
LGTM

PR #255, opened 2024-05-11
feat(moderation_request): Added current project under moderation request

amritkv:
Added current project under moderation request detail page


hoangnt2 (approved):
Look good to me

PR #256, opened 2024-05-13
fix(moderation_request): Fixed http status condition for last moderator

amritkv:
Fixed http status condition for last moderator in unassign moderator at moderation request page


heliocastro (approved):
LGTM

PR #257, opened 2024-05-13
fix(edit_project): Added missing toast message on click of edit project button

amritkv:
fix(edit_project): Added missing toast message on click of edit project button

1. If the logged in user has created the project then the toast message should be displayed as "You are editing the original document" 
2. If not then it will display as "You will create a moderation request if you update"


heliocastro (approved):
LGTM

PR #258, opened 2024-05-13
fix(update_project): Fixed toast message on click of update project button

amritkv:
Added proper toast message on click of update project button


heliocastro (approved):
LGTM

PR #259, opened 2024-05-13
fix(edit_project): Fixed translations and button header title at edit project page

amritkv:
Fixed translations and button header title (i.e. project_name (project_version)) at edit project page


heliocastro (approved):
LGTM

PR #260, opened 2024-05-13
fix(project_detail): Fixed translations at project detail page

amritkv:
Fixed missing translations at project detail page


heliocastro (approved):
LGTM

PR #263, opened 2024-05-16
fix(moderation_request): Fixed moderation request count in button header title

amritkv:
Fixed moderation request count in button header title in the format below :
`MODERATIONS (Open moderation request count / Closed moderation request count)`


heliocastro (approved):
LGTM

PR #264, opened 2024-05-20
feat(spdx): Define Object for SPDX Document

hoangnt2:
This PR does:
- Define objects that are used by SPDX document feature.
- Define new env variable to enable SPDX document feature: **NEXT_PUBLIC_SPDX_ENABLE** (can be true | false)


heliocastro (commented):
@hoangnt2 Have some questions:
- Don't we have an typescript SPDX library yet. I was wondering.
- Let SPDX be enabled by default. We need sit us one day to decide how we will set a modular system to turn on and off features not only including this, but per type of user level. Is a thing for next meetings.

hoangnt2:
> * Don't we have an typescript SPDX library yet. I was wondering.

I don't think we need need an typescript SPDX library. Everything will be handled on the server side

> * Let SPDX be enabled by default. We need sit us one day to decide how we will set a modular system to turn on and off features not only including this, but per type of user level. Is a thing for next meetings.

I am wondering about this too. I also have some questions:

1. How we can make the backend and frontend configuration consistently?
- Currently, sw360 backend options will be located in /etc/sw360/sw360.properties (for **SPDX document** feature it will is spdx.document.enabled = true)
- In this PR, I created an environment **NEXT_PUBLIC_SPDX_ENABLE** to enable this feature on frontend side. But I think having to configure both the backend and frontend is not a good idea.

=> My idea is creating an endpoint on sw360 backend. It will provide the configurations that are setted on it. Then sw360-frontend will read the configurations from that endpoint. We can cache them to reduce requests to server.

2. Do we need a GUI where user can change the configurations?

heliocastro:
> => My idea is creating an endpoint on sw360 backend. It will provide the configurations that are set on server. Then sw360-frontend will read the configurations from that endpoint. We can cache the response to reduce requests to server.
> 
> 2. Do we need a GUI where user can change the configurations?
@hoangnt2 This is exactly what i was thinking. We request a single login time the configuration from backend, and we can setup the menu, the modules, everything following the user  / group capabilities. 
The yes, we will need a Gui for that. But one step a time. I don't think we can pull this off for the first release.
I'm approving this PR.
Can you please add this as a discussion ?

PR #270, opened 2024-05-22
feat(CR): Added open CR and CR details page

amritkv:
- Landing page with fetched data in open CR table
- Redirection from open CR to CR detail page (with CR details and CR decision table)


heliocastro (approved):
Very good

PR #274, opened 2024-05-27
fix(CR): Fixed open CR list table data based on CR status

amritkv:
Fixed open CR list table data based on CR status

Out of 10 status list, if the CR status is `REJECTED` and `CLOSED`, it will go to Closed CR list. Rest will be in Open CR list


heliocastro (approved):
Looks good

PR #289, opened 2024-07-15
feat(Project): Implement project dependency network view (dependency network feature)

hoangnt2:
Implement project dependency network view (dependency network feature):
- Add new env variable to enable/disable the feature
- Show tree view + list view of dependency network in license clearing tab
- Fix TreeTable component that allow fetch children node when click expand button


heliocastro (approved):
LGTM

PR #292, opened 2024-07-22
feat(CR): Added scenario to handle deleted project after CR creation

amritkv:
Handle CR details and Open CR table if project is deleted after CR creation :

Open CR Table : 
- Show `Deleted Project` in open CR table.
- Show releases as `not available`

CR Details page :
- Disable edit CR button if project is deleted
- Change the CR details header for deleted project


heliocastro (approved):
LGTM

PR #296, opened 2024-07-28
fix(CR): Code refactoring for open CR table and open CR detail page

amritkv:
Code refactoring for open CR table and open CR detail page


heliocastro (approved):
LGTM

PR #297, opened 2024-07-29
feat(packages): Add package details page

deo002:
### Url
http://localhost:3000/packages/detail/{id}

### Screenshots
image


src/app/[locale]/packages/detail/[id]/detail.module.css
Comment (heliocastro) [inline src/app/[locale]/packages/detail/[id]/detail.module.css:1]: Please do not reinvent the wheel.
Why add **again** css for components that are global.
Please reuse the global css.

heliocastro (changes requested):
Remove custom css and use global on

deo002:
The css was used locally at many places. Making it global required changes in a lot of places which made the PR big. Should I add the second commit in another PR? @heliocastro

heliocastro:
> The css was used locally at many places. Making it global required changes in a lot of places which made the PR big. Should I add the second commit in another PR? @heliocastro

Yes, you should.
The fact is using in a lot places is enough information that should be global. Local solution just hurt maintenance, quality of the code, and future possible changes

deo002:
> > The css was used locally at many places. Making it global required changes in a lot of places which made the PR big. Should I add the second commit in another PR? @heliocastro
> 
> Yes, you should. The fact is using in a lot places is enough information that should be global. Local solution just hurt maintenance, quality of the code, and future possible changes

Ok, I'll add the follow up PR after rebasing once this one is merged.

heliocastro (approved):
I will merge this one. 
@deo002 Please fix this. I will remember you from now one every time you make a new PR, until this is fixed.
@hoangnt2 We should do a cleanup on CSS, and after that forbid any of this things again.

heliocastro:
@deo002 You need rebase this commit

deo002:
> I will merge this one. @deo002 Please fix this. I will remember you from now one every time you make a new PR, until this is fixed. @hoangnt2 We should do a cleanup on CSS, and after that forbid any of this things again.

I'll push a fix right after this is merged

heliocastro (approved):
This CSS need to move to global...

PR #299, opened 2024-08-05
feat(Create_CR): Added feature to change preferred clearing date based on CR priority

amritkv:
Added feature to change preferred clearing date based on CR priority

- By default the requested clearing date will be T+d days later.
- But if `Critical` flag is checked while creating CR, user can select requested clearing date from T day itself.

Issue : https://github.com/eclipse-sw360/sw360-frontend/issues/267


hoangnt2 (approved):
Look good to me

PR #300, opened 2024-08-06
feat(View_CR): Added view CR details at license clearing of project detail page

amritkv:
View CR:
- If a CR is in `OPEN` state : It will show `View Clearing Request` button and on click to it, it will show the existing CR details in a popup
- If a CR is in `CLOSED` state : It will disable the `Create Clearing Request` button.
- If a CR is in `IN_PROGRESS` state : It will show `Create Clearing Request` button and on click to it, it will ask the details to create a CR in a popup.

### DIFF
--- a/src/app/[locale]/projects/detail/[id]/components/ViewClearingRequestModal.tsx
+++ b/src/app/[locale]/projects/detail/[id]/components/ViewClearingRequestModal.tsx
@@ -12,7 +12,7 @@
 import { signOut, useSession } from "next-auth/react"
 import { useTranslations } from "next-intl"
 import { Dispatch, SetStateAction, useCallback, useEffect, useState } from "react"
-import { Button, Form, Modal } from "react-bootstrap"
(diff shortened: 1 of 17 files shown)


src/app/[locale]/projects/detail/[id]/components/ViewClearingRequestModal.tsx
```diff
@@ -0,0 +1,214 @@
...
+            } else if (response.status == HttpStatus.UNAUTHORIZED) {
+                return signOut()
+            } else {
+                notFound()
+            }
+        },[session]
+    )
+
+    useEffect(() => {
+        void fetchData(`clearingrequest/${clearingRequestId}`).then(
```
Comment (hoangnt2) [inline src/app/[locale]/projects/detail/[id]/components/ViewClearingRequestModal.tsx:86]: should check session is not null or undefined before fetching data

Comment (amritkv) [reply src/app/[locale]/projects/detail/[id]/components/ViewClearingRequestModal.tsx]: Addressed the issue.

hoangnt2 (approved):
Look good to me

PR #302, opened 2024-08-07
feat(Open_CR): Added edit button in open CR table

amritkv:
Added edit button in open CR table
- If the project is not deleted, the button will show up and take it to edit CR page
- If the project is deleted, the edit button will not be shown in open CR table


heliocastro (approved):
LGTM

PR #316, opened 2024-08-25
fix(CR_Details): Added request closed on field in closed CR detail page

amritkv:
Added request closed on field in closed CR detail page


heliocastro (approved):
LGTM

PR #320, opened 2024-09-12
feat(CR): Added clearing request comments tab

amritkv:
Added clearing request comments tab
- Lazy loading for comments
- formatted comments as per existing UI
- Add  new comments feature


src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx
```diff
@@ -10,24 +10,108 @@
...
+        if (response.status == HttpStatus.OK) {
+            const response_data = await response.json() as EmbeddedClearingRequestComments
+            setInputComment('')
+            setComments(response_data._embedded['sw360:comments'])
+            MessageService.success(t('Your comments updated successfully'))
+        } else if (response.status == HttpStatus.UNAUTHORIZED) {
+            return signOut()
+        } else {
+            MessageService.error(t('There are some problem to update your comments'))
+            notFound()
```
Comment (hoangnt2) [inline src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx:105]: I think we don't need notFound here

src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx
```diff
@@ -10,24 +10,108 @@
...
-    const { status } = useSession()
+    const [loading, setLoading] = useState(true)
+    const { data:session, status } = useSession()
+    const [comments, setComments] = useState<Array<ClearingRequestComments>>([])
+    const [inputComment, setInputComment] = useState('')
+    const [commentPayload, setCommentPayload] = useState({
+        text: ''
+    })
+
+    const formatDate = (timestamp: number): string | null => {
```
Comment (hoangnt2) [inline src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx:38]: I think this function can be re-write as below
```
const formatDate = (timestamp: number): string | null => {
        if (!timestamp) {
            return null
        }
        const date = new Date(timestamp)
        const dateISOString = date.toISOString()
        return `${dateISOString.slice(0,10)}  ${dateISOString.slice(11,19)}`
    }
```

src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx
Comment (amritkv) [reply src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx]: Thanks @hoangnt2 for pointing out. I forgot to remove it.
Now it is fixed ! 👍🏻

src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx
Comment (amritkv) [reply src/app/[locale]/requests/clearingRequest/detail/[id]/components/ClearingComments.tsx:38]: @hoangnt2, I have refactored it.

hoangnt2 (approved):
Look good to me

PR #328, opened 2024-09-23
fix(Style): Makes the styling of index pages consistent

hoangnt2:
Makes the styling of index pages consistent


heliocastro (approved):
LGTM

PR #334, opened 2024-10-02
fix(Code): Fix linter error for Home and Auth page

hoangnt2:
Fix linter error for Home and Auth page


heliocastro (approved):
Always good @hoangnt2

PR #337, opened 2024-10-04
fix: linter for preferences

duonglq-tsdv:


heliocastro (approved):
LGTM
Welcome on board

PR #342, opened 2024-10-16
fix(Fetch): Fix bug fetch data return 404 error

hoangnt2:
- Fix bug fetch data return 404 error at Component, Project, Release pages
- Fix bug search release modal not working


heliocastro (approved):
LGTM

PR #349, opened 2024-10-31
Fix project dependency network off mode

hoangnt2:
- Correct add/update/duplicate project when dependency network is disabled
- Handle null and undefined case for convertObjectToMapRoles


heliocastro (approved):
LGTM

PR #350, opened 2024-10-31
Fix linter for global components (part 1)

hoangnt2:
This PR fix linter for some global components. Other components will be fixed in next PRs


heliocastro (approved):
Looks good. Almost there.

PR #359, opened 2024-11-29
Fix bugs for detail/edit release page

hoangnt2:
- Fix bug cannot show release detail page because of undefined PackageVerificationCode
- Fix bug cannot update release when **NEXT_PUBLIC_SPDX_ENABLE=false**
- Handle create moderation request for non ADMIN user


heliocastro (approved):
Good to go

PR #394, opened 2024-12-10
fix(SearchVendors): Prevent fetching vendors on component initialization

hoangnt2:
### Bug:
Vendors are fetched when the component is initialized.

### Fix:
Fetch vendors only when the user clicks on the Search button.

image


heliocastro (approved):
Good to go

PR #408, opened 2024-12-16
refactor(requests): Refactor requests folder

deo002:


src/app/[locale]/requests/moderationRequest/[id]/components/currentRelease/CurrentReleaseDetail.tsx
```diff
@@ -51,26 +51,30 @@ const CurrentReleaseDetail = ({ releaseId }: Props): JSX.Element => {
...
-            if (response.status == HttpStatus.OK) {
-                const data = await response.json()
-                return data
-            } else if (response.status == HttpStatus.UNAUTHORIZED) {
-                return signOut()
-            } else {
-                return null
+        async (url: string, signal: AbortSignal) => {
+            try {
+                if (CommonUtils.isNullOrUndefined(session)) return signOut()
```
Comment (hoangnt2) [inline src/app/[locale]/requests/moderationRequest/[id]/components/currentRelease/CurrentReleaseDetail.tsx:56]: I know you are working on this PR, but I have a comment. If you use useSession(), there is a delay (loading time) before the session is retrieved. So, if you call this fetch function inside useEffect() and the session has not been retrieved yet, the user might be logged out. I think you should use getSession() instead to avoid that problem.

Comment (heliocastro) [reply src/app/[locale]/requests/moderationRequest/[id]/components/currentRelease/CurrentReleaseDetail.tsx]: @deo002 Have you addressed @hoangnt2 concerns ?

Comment (deo002) [reply src/app/[locale]/requests/moderationRequest/[id]/components/currentRelease/CurrentReleaseDetail.tsx]: Yes, I have. @heliocastro

heliocastro (approved):
Issues are addressed

PR #409, opened 2024-12-16
refactor: Move to pnpm

heliocastro:
Move the main package manager to pnpm


heliocastro:
@hoangnt2 The change for pnpm has an interesting side effect, is running cypress tests by default.

hoangnt2 (approved):
Look good to me

hoangnt2:
@heliocastro Sorry, but what do you mean by 'is running Cypress tests by default'?

heliocastro:
> @heliocastro we are using e2e testing strategy for cypress, so a running application is needed for testing. Running cypress while building application will cause it to error

No no, build check the cypress tests, if all is ready to run.

PR #411, opened 2024-12-19
feat(Project): Added default license info header in add project page

amritkv:
Added default license info header in add project page


heliocastro (approved):
Good to go

PR #414, opened 2024-12-20
fix(Component): Fix bug component owner is not showing on component detail page

hoangnt2:
Fix bug component owner is not showing on component detail page
image


heliocastro (approved):
LGTM

PR #425, opened 2025-01-02
fix(Project): Fixed the alignment of default license info header button

amritkv:
Fixed the alignment of default license info header button


heliocastro (approved):
Good to go ( We should use more tailwind classes )

PR #438, opened 2025-01-15
fix(Project): Fixed view and create CR at project page

amritkv:
Fixed view and create CR details at project page


hoangnt2 (approved):
Looks good to me

PR #439, opened 2025-01-17
Fixed Visibility field in Project detail

duonglq-tsdv:
Fixed the Visibility field in Project detail page to display **"Group And Moderators" i**nstead of **"Buisnessunit And Moderators"**.


heliocastro (approved):
Good to go

PR #472, opened 2025-02-07
feat(AdvancedSearch): Submit search on Enter

afsahsyeda:
**Description**
Added onKeyDown event listener for the string fields of Advanced search to call the onSubmit function when the pressed key is Enter.
This makes the user experience more intuitive as they they don't have to scroll to the bottom to press the "Submit" button.


hoangnt2 (commented):
Looks good

heliocastro (approved):
LGTM

PR #504, opened 2025-02-28
fix(components): Fix the href url that will cause 404

pominc0624:
Summary: 
---
This pull request fixes the issue where clicking the linked release in the Components section results in a 404 error.

Problem:
--- 
The URL for the linked release was incorrect, causing a 404 error when clicked. 

For example: 

The error URL was:
http://localhost:3000/components/releases/detail/components/releases/details/7b7c30a808442dc7aed54e093303c031

The correct URL should be:
http://localhost:3000/components/releases/detail/7b7c30a808442dc7aed54e093303c031

This fix updates the URL to the correct format, ensuring that the linked release can be accessed without any errors.


heliocastro (approved):
Good to go

PR #515, opened 2025-03-06
fix(licenses): Fix the license url that cause 404

pominc0624:
Summary: 
---
This pull request fixes the issue where clicking the license detail in the Search Bar and the components table that cause 404 error.

This fix updates the URL to the correct format, ensuring that the license detail can be accessed without any errors.


heliocastro (approved):
Good to go.

Issue #607, opened 2025-03-12
NextJS 15.2.1 error


hoangnt2:
This error occurs in build mode  (pnpm start)
![Image](https://github.com/user-attachments/assets/edc7223d-8a19-40b0-a642-546a84a89661)

Nextjs issue: https://github.com/vercel/next.js/issues/76817

amritkv:
PR Merged #608 !

PR #608, opened 2025-03-12
fix(NextJS): Upgrade NextJS to v15.2.2 to fix setting 'SHA224' error

hoangnt2:
Issue: #607


hoangnt2:
@heliocastro, the front-end is currently not working. We need to merge this PR as soon as possible.

heliocastro (approved):
Chore

heliocastro:
@hoangnt2 I think this cases that are more urgent, you and me are ok to self commit. It hurts a little the scorecard, but is better then have project unstable

PR #616, opened 2025-03-16
fix(Home):Navigation Does Not Maintain Locale Path When Redirecting

### LINKED ISSUE #615: Home: Navigation Does Not Maintain Locale Path When Redirecting
## Router.push function doesnot maintain its localized paths
I would like to workon this issue.

VAIBHAVSING:
fixes #615


heliocastro (approved):
Good to go

Issue #624, opened 2025-03-17
Build Failure: TypeScript Error in RefObject Definition in TableLicense Component


harshitg927:
# Description
When building the project, there's a TypeScript error in the TableLicense.tsx component related to the ref attribute:

./src/components/LinkedObligations/TableLicense.tsx:129:22
Type error: Type 'RefObject<HTMLDivElement | null>' is not assignable to type 'LegacyRef<HTMLDivElement> | undefined'.
  Type 'RefObject<HTMLDivElement | null>' is not assignable to type 'RefObject<HTMLDivElement>'.
    Type 'HTMLDivElement | null' is not assignable to type 'HTMLDivElement'.
      Type 'null' is not assignable to type 'HTMLDivElement'.
  127 |                     )}
  128 |                 </div>
> 129 |                 <div ref={this.wrapper} />
      |                      ^
  130 |             </>
  131 |         )
  132 |     }

## Problem 
The wrapper ref is defined with a type that includes null explicitly in the union type:
private wrapper: RefObject<HTMLDivElement | null> = createRef()
However, React's ref attribute expects a RefObject<HTMLDivElement> without the extra null type.

## Solution
Remove the | null from the type definition:
private wrapper: RefObject<HTMLDivElement> = createRef()
This change maintains the intended functionality since createRef() already returns a RefObject that can hold null internally (as its .current property). The explicit | null in the type was causing a type mismatch with React's expected ref types.

amritkv:
Closing this issue, as this is not relevant anymore.

PR #625, opened 2025-03-17
Fix: TypeScript build failure in TableLicense component's RefObject definition

harshitg927:
# Description
This PR fixes a build failure caused by a TypeScript error in the TableLicense.tsx component. The error occurs because the wrapper ref is defined with an explicit null in its union type, which conflicts with React's expected ref types.

# Changes
 - Modified the type definition of wrapper ref in TableLicense.tsx and Table.tsx from RefObject<HTMLDivElement | null> to RefObject<HTMLDivElement>

# Issue
Fixes: [#624](https://github.com/eclipse-sw360/sw360-frontend/issues/624)

### DIFF
--- a/src/components/LinkedObligations/TableLicense.tsx
+++ b/src/components/LinkedObligations/TableLicense.tsx
@@ -34,7 +34,7 @@ interface TableProps extends Partial<Config> {
 }
 
 class TableLicense extends Component<TableProps, unknown> {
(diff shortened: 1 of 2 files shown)


heliocastro (approved):
Good to go

hoangnt2:
Hi @harshitg927 , I fixed build error in https://github.com/eclipse-sw360/sw360-frontend/pull/650.

PR #638, opened 2025-03-21
fix(config): Fixed ERR_REQUIRE_ESM error while creating commits

amritkv:
Fixed this issue of `ERR_REQUIRE_ESM` while creating commits, by upgrading `commitlint` package version.


heliocastro (approved):
Good to go

PR #699, opened 2025-04-23
feat(ECC) : Enhancement of the release "ECC Details" view

bibhuti230185:
This is a boolean value that only ECC experts are allowed to set.\
The field is to document that an ECC expert decided that a certain release contains cryptography in a way that it might be relevant for the ECC classification of a product that uses this release.

Issue: 
https://github.com/eclipse-sw360/sw360/issues/3102

### Suggest Reviewer
@GMishx 

### How To Test?
image

image


src/app/[locale]/components/releases/detail/[id]/components/ECCDetails.tsx
```diff
@@ -59,6 +59,16 @@ const ECCDetails = ({ release }: Props) : ReactNode => {
...
                     </tr>
+                    <tr>
+                        <td>{t('Contains Cryptography')}:</td>
+                        <td>
+                            {release.eccInformation?.containsCryptography === true
+                                ? t('Yes')
+                                : release.eccInformation?.containsCryptography === false
+                                ? t('No')
+                                : t('Not Specified')}
+                        </td>
```
Comment (heliocastro) [inline src/app/[locale]/components/releases/detail/[id]/components/ECCDetails.tsx:70]: Are you sure this logic works ?

heliocastro (changes requested):
Make this a classic if / else because that logic is npredictable

src/app/[locale]/components/releases/detail/[id]/components/ECCDetails.tsx
Comment (bibhuti230185) [reply src/app/[locale]/components/releases/detail/[id]/components/ECCDetails.tsx:70]: @heliocastro Thanks for the review. I have cross verified the above condition and it's working fine.
image
image
image

heliocastro (approved):
Good to go

PR #716, opened 2025-04-28
Merge Components Section

deo002:
Merge components Merge Confirmation page and left over Merge Data page completed.


src/app/[locale]/components/detail/[id]/merge/merge.module.css
```diff
@@ -29,4 +29,8 @@ License-Filename: LICENSE
 
 .green {
     color: green;
+}
+
+.orange {
+    background-color: #fff4ec;
```
Comment (heliocastro) [inline src/app/[locale]/components/detail/[id]/merge/merge.module.css:35]: @deo002  Can you look on this css module and remove the duplications and start to use global css ?
If we change the layout of global CSS, this will break

heliocastro (changes requested):
Please take a look on the CSS possible heasdache

src/app/[locale]/components/detail/[id]/merge/merge.module.css
Comment (deo002) [reply src/app/[locale]/components/detail/[id]/merge/merge.module.css]: @heliocastro yes, will do it

deo002:
While testing please use the latest main. PR https://github.com/eclipse-sw360/sw360/pull/3117 solves a bug that causes the database to have some fields missing.

PR #717, opened 2025-04-29
feat(Admin): Added SW360_ADMIN role to access admin tab

amritkv:
Added SW360_ADMIN role to access admin tab.
Closes issue : https://github.com/eclipse-sw360/sw360-frontend/issues/691


GMishx (approved):
Changes looks good.

PR #755, opened 2025-05-22
fix(translations): Fixed project wide translations

amritkv:
Fixed project wide translations

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/754


heliocastro (approved):
Good to Go

PR #763, opened 2025-05-29
feat(config): Add forms for domain regex config

GMishx:
Add UI elements for new config `release.sourcecodeurl.skip.domains` added by https://github.com/eclipse-sw360/sw360/pull/3177


hoangnt2 (approved):
Look good to me

PR #774, opened 2025-06-02
fix(licenses): Fix uncheck alert and other errors

### LINKED ISSUE #690: Fix checked status at license detail page
The checked status should be same as per below images. But as of now it's coming `UNCHECKED` always.

Image

Image

deo002:
Closes #690


hoangnt2 (approved):
Code looks good to me

PR #788, opened 2025-06-09
refactor: Use proper eslint 9.x setup with flat config

heliocastro:
- Updated to Flat Config mjs eslint config
- Move all possible configurations to eslint and package.json
- Reformatted all src files to match prettier definitions. Prettier settings lives in package.json now
- Turned on as warnings all Typescript checks to made aware


heliocastro:
@hoangnt2 @amritkv Since we are approaching the first release preview. I turned on new eslint checks, and  formatted all files to the defined prettier entries.

For typescript errors, I moved to warnings, but we do have +100 real issues.

For code layout, now prettier is marked as error, so no more commits with invalid layout.

I will pause all other commits until you two review the eslint and package.json parts. ( the src is toomuch )

PR #808, opened 2025-06-23
feat(Release): Create MR based on user role to update Release

amritkv:
Create MR based on user role to update Release.
Closes issue : https://github.com/eclipse-sw360/sw360-frontend/issues/792


heliocastro (dismissed):
Good to go

amritkv:
Hey @heliocastro !
It's not a duplicate. But in some sense, you can say so.
I mean while updating a `Project`, `Component` or `Release`, based on user's role, if MR is required, we need to pass `Comment` on it.
So, I've made the comment modal as generic and calling this modal from `Edit Project`, `Edit Component` and `Edit Release`.

PR #815, opened 2025-06-25
fix(Project): Fixed edit release url at project license clearing table

amritkv:
Fixed edit release url at project license clearing overview table


GMishx (approved):
Looks good.

PR #850, opened 2025-07-02
fix(Component): Removed group field from advance search

amritkv:
Removed group field from advance search


heliocastro (approved):
Good to go

Issue #853, opened 2025-07-03
Admin : Add missing fields at `User` page


amritkv:
Add missing fields at `User` page

Issue #855, opened 2025-07-03
Admin : Add `Email` at button header under `User` page


amritkv:
Add missing e-mail at button header i.e. top right corner of user edit page.

![Image](https://github.com/user-attachments/assets/b722b5cf-ad20-448a-a506-c6e2d5b71120)

amritkv:
PR merged : https://github.com/eclipse-sw360/sw360-frontend/pull/856

Issue #854, opened 2025-07-03
Admin : Add global identifier at `User` page


amritkv:
Add `ExternalId` at `User Details` and `User Edit` page.

PR #856, opened 2025-07-03
Admin : Fixed missing fields at `User` page

amritkv:
- Added global identifier at `User Detail` and `User Edit` page
- Added email at top right corner at `Edit User` page

Closes : #853 #854 #855


heliocastro (approved):
Good to go

Issue #879, opened 2025-07-17
Advanced Search : Fix `Exact Match` and `Orphan Package` warnings


amritkv:
Fix warnings on `Exact Match` and `Orphan Package` at `Advanced Search` component

amritkv:
PR merged : #880

Issue #837, opened 2025-06-30
Users Page - Search calls wrong endpoint


amritkv:
Performing Advance Search calls `GET` `/admin/users?allDetails=true&luceneSearch=true&email=**@abc.com` instead of `GET` `/resource/api/admin/users?allDetails=true&luceneSearch=true&email=**@abc.com`

PR #880, opened 2025-07-17
fix(Users): Fixed advanced search at users page

amritkv:
- Fixed advanced search at users page
- Fixed enable/disable of `Exact Match` and `Orphan Package` filter at `Advanced Search`

Closes : [#837](https://github.com/eclipse-sw360/sw360-frontend/issues/837)  [#879 ](https://github.com/eclipse-sw360/sw360-frontend/issues/879)


heliocastro (approved):
Good to go

PR #881, opened 2025-07-18
feat(tanstack): Migration from gridjs to tanstack table for projects table

deo002:


heliocastro (dismissed):
@deo002 You opened the flood gate. So now we won't release until all the tables are migrated.

heliocastro:
@deo002 And your branch still have conflicts

GMishx:
> @deo002 You opened the flood gate. So now we won't release until all the tables are migrated.

Unfortunately yes, without the new table, the sorting won't work and UI would be unusable for many use-cases anyways.

In terms of timeline, if we can finish the migration in a month, it should okay. Anyways, the new table is reported to have better performance and will be testing it on our staging instance soon with tons of data to render.

heliocastro (approved):
Cross fingers for next ones

PR #883, opened 2025-07-18
fix(Licenses): Added missing `License Type` at `License` detail page

amritkv:
Added missing data of `License Type` at `License` detail page

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/882


heliocastro (approved):
Good to go

PR #885, opened 2025-07-18
fix(Licenses): Fixed delete license with proper message

amritkv:
On delete license, the response will show proper messages.
- If the license in use, it will show on the popup : `Unable to delete license. License is in Use`
- Minor alignment and appearance updates

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/884


heliocastro (approved):
Good to go

PR #948, opened 2025-07-31
fix(SW360): Fixed `singout` functionality 

amritkv:
As of now, only at `Home` page, it checks for status of session and logs out based on token expiry. But now the token validity should be checked at all pages, modals and other components.

- Signout handled at `API` calls
- Signout handled at all pages, components and modals under `Projects` tab


heliocastro (changes requested):
- Need resolve conflitcs
- Need fix PR title

Issue #960, opened 2025-08-06
Create CR button visible in license clearing tab even though CR exists for the project.


akshitjoshii:
Currently if CR has been raised for a project but still in the license clearing tab we see **create CR** button instead of **view CR**

<img width="1190" height="212" alt="Image" src="https://github.com/user-attachments/assets/6a039a08-9ac6-4665-98d6-74c6a0ce501c" />

PR #970, opened 2025-08-11
fix(CR): Added fix for clearing request view.

akshitjoshii:
Closes : #960 

Currently the view CR modal is not being picked up in both License Clearing and Projects tab even if CR exists for a project.

Rebase with latest backend main for testing.

image


heliocastro (approved):
Good to go

PR #972, opened 2025-08-14
feat(vuln): Complete the implementation

### LINKED ISSUE #834: Vulnerabilities : Related Releases table is missing
Current : 

Image

Expected : 

Image

GMishx:
This PR adds more elements to the Vulnerabilities pages and bring it closer to completion.

Added components:
1. List of releases having the vulnerability.
2. Metadata tab of Vulnerability.
3. References tab of Vulnerability.

Fix following displays:
1. The vulnerability can have `lastExternalUpdate` or `lastUpdateDate` defined. Use one.
2. `var.substring(0, var.lastIndexOf('T'))` does not work if the variable does not contain `T` which causes `lastIndexOf()` to return `-1` and `substring(0, -1)` becomes `(-1, 0)` returning an empty string. This has been simplified to `var.substring(0, 10)` as the time is in format `YYYY-MM-DDTHH:MM:SS.zzz` where date is guaranteed to end in first 10 characters.

Created function:
1. `DisplayMapOfMap` which renders `Map<string, Map<string, string>>`
2. `DisplayMap` to render `[string, string][]`

Closes #834


heliocastro (approved):
Good to go

PR #1039, opened 2025-09-26
fix(Admin): Restrict admin sub page access to non admin roles

amritkv:
Restricted the admin sub-page access to non admin user roles

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/1038


heliocastro (commented):
@amritkv, is it not better to restrict pages by user groups?
This sound really flaky

PR #1042, opened 2025-09-27
fix(CR): Highlight open CRs in projects tab

akshitjoshii:
Highlight projects with open CRs in the projects tab.
image

### DIFF
--- a/src/app/[locale]/projects/components/Projects.tsx
+++ b/src/app/[locale]/projects/components/Projects.tsx
@@ -199,6 +199,8 @@ function Project(): JSX.Element {
                 cell: ({ row }) => {
                     const id = row.original['_links']['self']['href'].split('/').at(-1)
                     const projectClearingRequestId = row.original.clearingRequestId
+                    const clearingState = row.original.clearingState
+                    const isOpenClearingRequest = clearingState === 'OPEN'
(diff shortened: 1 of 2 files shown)


src/app/[locale]/projects/components/Projects.tsx
```diff
@@ -222,7 +224,8 @@ function Project(): JSX.Element {
                                             >
                                                 <MdOutlineTask
                                                     size={25}
-                                                    className='btn-icon overlay-trigger'
+                                                    className={`btn-icon overlay-trigger ${isOpenClearingRequest ? 'text-warning' : ''}`}
+                                                    style={isOpenClearingRequest ? { color: '#ffc107' } : {}}
```
Comment (heliocastro) [inline src/app/[locale]/projects/components/Projects.tsx:228]: No hardcoded colors.
Update CSS properly

heliocastro (changes requested):
Please avoid hardcoded information that should go to css.
The global CSS before somenon try to create another local one.

src/app/[locale]/projects/components/Projects.tsx
Comment (akshitjoshii) [reply src/app/[locale]/projects/components/Projects.tsx]: addressed

heliocastro (approved):
Good to go

PR #1061, opened 2025-10-03
fix(User): Fixed user search based on input text

amritkv:
Fixed users' search based on input keyword on Search Users' modal.

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/1060


heliocastro (approved):
Good to go

PR #1109, opened 2025-10-15
feat(types): Migrate to http-status-codes

heliocastro:


src/app/[locale]/admin/departments/components/SecondaryDepartmentsTable.tsx
```diff
@@ -32,9 +33,18 @@ const SecondaryDepartmentsTable = (): JSX.Element => {
...
+    ])
+
+    const columns = useMemo<
+        ColumnDef<
+            [
+                string,
+                string[],
+            ]
+        >[]
+    >(
```
Comment (amritkv) [inline src/app/[locale]/admin/departments/components/SecondaryDepartmentsTable.tsx]: Hey @heliocastro !
In my opinion, these auto formatted codes with un-necessary extra lines are making it difficult to read the code.
I've seen a lot of places the code is auto formatted like this (with extra lines)

Comment (heliocastro) [reply src/app/[locale]/admin/departments/components/SecondaryDepartmentsTable.tsx]: Yes and no. If you are used to seeing everything inline, it's ok, but then I can easily find lines that grow longer than our hard 120-character limit, and the inline becomes a mess.
And it's hard to make it consistent, since nobody agrees on when to break the line.
That's why I push this default and intend for you to finish the table work to make this a hard QA on CI.

PR #1173, opened 2025-11-03
Fix/copy to clipboard in Project page

Suhas2109:
Fixed the issue where the “Copy to Clipboard” tooltip didn’t change to “Copied” when clicked.
The tooltip now properly updates to “Copied” on click.


amritkv (changes requested):

Suhas2109:
Thanks @amritkv! I’ve removed the duplicate navigator.clipboard.writeText call and updated the click handler to use await as suggested.

PR #1177, opened 2025-11-06
fix(CR): Clearing Request creation is disabled wrongly for all projects

akshitjoshii:
Clearing request creation was blocked for all projects due to bug in the `isCrAllowed` function. Also added check for visibility of project.

image


heliocastro (approved):
Good to go

PR #1201, opened 2025-11-17
fix(message): Stick on top

GMishx:
Stick the message to top without increasing the height of the page.

The current alert message sticks on top by increases the height of the page causing the content to move.

Making the position to fixed prevents the content from moving.

Also, fix the search parameter name from `keyword` to `searchText` in licenses page.


heliocastro (approved):
Goof to go

PR #1218, opened 2025-11-19
feat(css): Add more classes to disabled cursors

GMishx:
Add `.form-control[readonly]` and `.form-control[disabled]` to list of elements where `cursor: not-allowed`


heliocastro (approved):
Good to go

PR #1220, opened 2025-11-20
fix(biome): Enable strict checks

heliocastro:


heliocastro:
> @heliocastro I just tried `pnpm biome check` and got
> 
> ```
> Found 522 errors.
> Found 44 warnings.
> ```

Yep, to see how long we have ahead..
Mostly i'm a little worried for the extensive usage of useeffect and how problematic ( performance and security wise ) this can be

PR #1283, opened 2025-12-23
Fix: Session caching after logout

### LINKED ISSUE #876: SW360 - Fix token fetching after logout
Once user logs out from ui and clicks back button, `/session` gives the keycloak session token

manhuu14:
Fixes #876

## Problem
Once user logs out from UI and clicks back button, `/session` still gives the Keycloak session token, which shouldn't happen after logout.

## Root Cause
1. Session endpoint was missing cache control headers, causing browsers to cache responses
2. Keycloak session was not being invalidated on logout

## Solution
1. Added cache control headers to `/session` endpoint to prevent browser caching
2. Added `signOut` event handler in NextAuth to properly invalidate Keycloak session

## Changes
- `src/app/api/session/route.ts`: Added cache control headers (`no-store, no-cache, must-revalidate, private`)
- `src/app/api/auth/[...nextauth]/keycloakAuthOption.ts`: Added `events.signOut` handler to invalidate Keycloak session

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.3.6/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.3.10/schema.json",
   "vcs": {
     "enabled": false,
     "clientKind": "git",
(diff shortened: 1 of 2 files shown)


manhuu14:
@heliocastro  due to technical issue code in my pc collapsed suddenly. and now i re-forked the repo and now created pr 
please review it once ThankYou!

heliocastro (approved):
Good to go

PR #1287, opened 2025-12-26
feat(components): Add autocomplete for Advanced Search fields in components page

taanvi2205:
- Add autocomplete suggestions for Languages, Software Platforms, Operating Systems
- Add autocomplete for Vendors using API
- Use existing SuggestionBox component and useConfigValue hook

Closes: eclipse-sw360/sw360-frontend#1286

### DIFF
--- a/src/app/[locale]/projects/detail/[id]/components/Summary.tsx
+++ b/src/app/[locale]/projects/detail/[id]/components/Summary.tsx
@@ -9,6 +9,7 @@
 
 'use client'
 
+import { decode } from 'he'
 import Link from 'next/link'
 import { signOut, useSession } from 'next-auth/react'
 import { useTranslations } from 'next-intl'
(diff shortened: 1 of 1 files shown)


src/app/[locale]/components/components/ComponentIndex.tsx
```diff
@@ -35,6 +41,26 @@ const ComponentIndex = (): ReactNode => {
...
+            try {
+                const response = await ApiUtils.GET('vendors', session?.user?.access_token || '')
+                if (response.status === 200) {
+                    const data = await response.json()
+                    const names = data._embedded?.['sw360:vendors']?.map((v: { fullName: string }) => v.fullName) || []
+                    setVendorsSuggestions(names)
+                }
+            } catch (error) {
+                console.error('Failed to fetch vendors:', error)
+            }
```
Comment (heliocastro) [inline src/app/[locale]/components/components/ComponentIndex.tsx]: No console messages or try catch inside useEffecgt

heliocastro (approved):
Good to go

taanvi2205:
Hi @heliocastro, it looks like the merge commit introduced some duplicate code duplicate controller declarations and return statements which is causing the Biome CI to fail but my local branch is passing all the tests. What should i do now?

PR #1303, opened 2025-12-31
feat(components): Add autocomplete for Advanced Search fields

heliocastro:
- Add autocomplete suggestions for Languages, Software Platforms, Operating Systems
- Add autocomplete for Vendors using API
- Use existing SuggestionBox component and useConfigValue hook

Closes: https://github.com/eclipse-sw360/sw360-frontend/pull/1287

Closes: eclipse-sw360/sw360-frontend#1286


heliocastro:
@taanvi2205 I decided commit your original code, but since the PR came from your space, i just fixed and add a secondary PR.
We will deal with the useeffect rende loop if become a problem.

Thanks for your commit.

heliocastro:
@taanvi2205 Made a small modication, your PR is going through.

PR #1315, opened 2026-01-06
feat: Add filters for releases without attachment usage set

### LINKED ISSUE #720: Feature request: Filter releases without usage set
With large projects, it is difficult to find which releases does not have License usage or source usage set. The idea is to filter out such releases in the UI and make it easy for user to find such releases and set the attachment usage for them.

The feature request is to create filter at 2 pages:

### 1. Attachment Usage
First part of the feature request is to add 2 new filters for "Attachment Usage" page.

Image

The filters can be called as "Without CLI usage set" and "Without source usage set". Setting one of these filter should remove all releases from the table except the releases which does not have CLX/License Info usage set or Source Code Bundle respectively for even single attachment.

### 2. Generate License Info & Generate Source Code Bundle pages
The second part of the feature request is to add a filter button on "Generate License Info" and "Generate Source Code Bundle" pages with similar effect.

Image

Idea is to create buttons as demonstrated above in "Generate License Info" page which removes all releases except the ones which does not have CLX attachment usage set for even one attachment.

Image

Similar button can be created in "Generate Source Code Bundle" page to remove all releases except the ones which does not have Source Code Bundle attachment usage set for even one attachment.

manhuu14:
## Description
Implements filtering functionality to help users identify releases that don't have proper attachment usage configured, addressing issue #720.

### 1. Attachment Usage Page
- Added two new filter options in the release filter dropdown:
  - "Without CLI Usage Set" - Shows releases with CLI/CLX/ISR attachments but no license info usage configured
  - "Without Source Usage Set" - Shows releases with SOURCE/SRC attachments but no source package usage configured
- Updated `releaseMatchesFilter` function to handle new filter cases
- Modified `filterRows` function to accept and use `saveUsagesPayload` for filtering
- Updated `filteredData` useMemo to include `saveUsagesPayload` dependency

### 2. Generate License Info Page
- Added "Hide Releases With Usage Set" / "Show All Releases" toggle button
- Implemented `hasCliUsageSet` helper function to check CLI attachment usage
- Updated `formatReleaseAttachmentDataToTableData` to respect the filter
- Updated `extractLinkedProjectsAndTheirLinkedReleases` to pass filter parameter
- Updated `buildTable` to accept and use `filterWithoutUsage` parameter
- Updated useEffect dependencies to include `hideWith 

Fixes #720

### DIFF
--- a/src/app/[locale]/licenses/detail/components/Obligations.tsx
+++ b/src/app/[locale]/licenses/detail/components/Obligations.tsx
@@ -11,13 +11,18 @@
 
 'use client'
 
-import { ColumnDef, getCoreRowModel, getExpandedRowModel, useReactTable } from '@tanstack/react-table'
+import {
+    ColumnDef,
+    getCoreRowModel,
+    getExpandedRowModel,
+    getFilteredRowModel,
+    useReactTable,
+} from '@tanstack/react-table'
 import { StatusCodes } from 'http-status-codes'
-import { useSearchParams } from 'next/navigation'
 import { signOut, useSession } from 'next-auth/react'
 import { useTranslations } from 'next-intl'
 import { PaddedCell, SW360Table } from 'next-sw360'
-import { ReactNode, useEffect, useMemo, useState } from 'react'
+import { ReactNode, useCallback, useEffect, useMemo, useState } from 'react'
 import { Form, Spinner } from 'react-bootstrap'
 import { ErrorDetails, LicenseDetail, NestedRows, Obligation } from '@/object-types'
 import MessageService from '@/services/message.service'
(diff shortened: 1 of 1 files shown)


deo002 (changes requested):
Thanks for your contributions! Suggesting some minor changes! Rest all looks good!

deo002 (dismissed):

manhuu14:
Implemented! Added the same filtering functionality to the Generate Source Code Bundle page:
- Added "Hide Releases With Usage Set" toggle button
- Implemented `hasSourceUsageSet` helper with exact string matching
- Updated all table building functions to support the filter

Both pages now have consistent filtering behavior.

PR #1317, opened 2026-01-07
fix(misc): Fix miscellaneous small bugs

deo002:
Link fixes, css fixes and handling undefined references


heliocastro (changes requested):
Nope nope nope nope
Split this in small commits.

deo002:
Done @heliocastro !

PR #1321, opened 2026-01-09
fix(component): Resolve delete conflict message placeholders

Suhas2109:
This change fixes an issue in the Delete Component dialog in the Component Portal where the conflict warning message displayed unresolved placeholder variables such as {name} and {releaseCount}.


akshitjoshii (approved):
LGTM

Issue #1322, opened 2026-01-09
Security: Credentials in URL Query Parameters


chauhan-varun:
## Severity
**CRITICAL** 🔴

## Location
**File:** `src/services/auth.service.ts`

```typescript
const authorizationURL: string =
  SW360_API_URL +
  '/authorization/oauth/token?grant_type=password&username=' +
  userData.username +
  '&password=' +
  userData.password
```

## Description
Passwords are passed in URL query parameters during authentication. This is a significant security vulnerability because:

1. **Server Logs** - URLs are typically logged by web servers, proxies, and load balancers, exposing credentials in plain text
2. **Browser History** - URLs with query parameters are stored in browser history
3. **Network Monitoring** - Query parameters are visible in network monitoring tools and can be intercepted
4. **Referrer Headers** - URLs may be leaked via HTTP Referrer headers when navigating to external sites
5. **Caching** - URLs may be cached by intermediate proxies or CDNs

## Recommendation
Use HTTP POST body for transmitting credentials instead of URL query parameters:

```typescript
const authorizationURL: string = SW360_API_URL + '/authorization/oauth/token';

const response = await fetch(authorizationURL, {
  method: 'POST',
  headers: {
    'Content-Type': 'application/x-www-form-urlencoded',
  },
  body: new URLSearchParams({
    grant_type: 'password',
    username: userData.username,
    password: userData.password,
  }),
});
```

## References
- [OWASP - Information Exposure Through Query Strings in URL](https://owasp.org/www-community/vulnerabilities/Information_exposure_through_query_strings_in_url)
- [CWE-598: Use of GET Request Method With Sensitive Query Strings](https://cwe.mitre.org/data/definitions/598.html)

PR #1323, opened 2026-01-09
fix(auth): Use POST and form data for token request

chauhan-varun:
[Screencast from 2026-01-09 19-33-33.webm](https://github.com/user-attachments/assets/b8a57106-67cc-41f5-bf76-0e8ae56f1e4e)
fixes: #1322

## Summary

This PR fixes a **critical security vulnerability** where user credentials (username and password) were being transmitted via URL query parameters during OAuth token requests.

This is a security risk because URL query parameters can be exposed in:
- **Server logs** - Web servers, proxies, and load balancers typically log full URLs
- **Browser history** - URLs with query parameters are stored in browser history
- **Network monitoring** - Query parameters are visible in network inspection tools
- **HTTP Referrer headers** - URLs may leak via Referrer headers when navigating externally
- **Proxy caches** - URLs may be cached by intermediate proxies or CDNs

## Testing

- [x] Manual login test to verify authentication still works
- [x] Verify network requests show credentials in body, not URL

### DIFF
--- a/src/services/auth.service.ts
+++ b/src/services/auth.service.ts
@@ -41,15 +41,22 @@ const generateToken = async (userData: UserCredentialInfo): Promise<null | AuthT
         credentials = Buffer.from(`${oAuthClient.client_id}:${oAuthClient.client_secret}`, `binary`).toString('base64')
 
         opts.headers['Authorization'] = `Basic ${credentials}`
-        const authorizationURL: string =
-            SW360_API_URL +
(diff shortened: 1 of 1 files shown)


heliocastro (approved):
Thanks for your contribution.
We do need improve the regular auth which is not recommended by default, only fr development, but there are use cases that users will not want a keycloack based auth.

PR #1335, opened 2026-01-13
feat(components): Show success notification on export spreadsheet action

Suhas2109:
This PR adds immediate user feedback when the Export Spreadsheet action is triggered in the Components portlet.

When a user selects Components only or Components with releases, a success notification is shown indicating that the spreadsheet is being prepared and will be emailed shortly.
- This change is UI-only
- Displays a green success notification immediately after the export action is triggered

image


amritkv:
Hey @Suhas2109 !
I think you need to change the status message a bit. This is because we are directly downloading the file in the browser and not sending any email notification to the user.

src/app/[locale]/components/components/ComponentIndex.tsx
```diff
@@ -205,6 +206,7 @@ const ComponentIndex = (): ReactNode => {
     ]
 
     const handleExportComponent = (withLinkedReleases: string) => {
+        setShowExportMessage(true)
```
Comment (amritkv) [inline src/app/[locale]/components/components/ComponentIndex.tsx]: Please capture the response code from `DownloadService.download` and based on the response code, show the message.
So, ideally the ` setShowExportMessage(true)` should be set after receiving response code from `DownloadService.download` call.

src/object-types/enums/UIConfigKeys.ts
```diff
@@ -31,6 +31,7 @@ export enum UIConfigKeys {
     UI_ENABLE_SECURITY_VULNERABILITY_MONITORING = 'ui.enable.security.vulnerability.monitoring',
     UI_REST_APITOKEN_WRITE_GENERATOR_ENABLE = 'ui.rest.apitoken.write.generator.enable',
     UI_SVM_NOTIFICATION_URL = 'svm.notification.url',
+    UI_EXPORT_SPREADSHEET_MAIL_REQUEST = 'ui.export.spreadsheet.mail.request',
```
amritkv: Hey @Suhas2109 !
It seems the config key is wrong.
Actual key is : `send.component.spreadsheet.export.to.mail.enabled`

src/app/[locale]/components/components/ComponentIndex.tsx
```diff
@@ -204,13 +207,20 @@ const ComponentIndex = (): ReactNode => {
...
         const currentDate = new Date().toISOString().split('T')[0]
-        DownloadService.download(
-            `reports?withlinkedreleases=${withLinkedReleases}&mimetype=xlsx&mailrequest=false&module=components`,
+        const response = await DownloadService.download(
+            `reports?withlinkedreleases=${withLinkedReleases}&mimetype=xlsx&mailrequest=${exportViaMail ?? false}&module=components`,
             session,
             `components-${currentDate}.xlsx`,
         )
+        if (response === 200) {
+            setShowExportMessage(true)
```
Comment (amritkv) [inline src/app/[locale]/components/components/ComponentIndex.tsx]: @Suhas2109 !
The banner message looks good but a small change is required.
If the email notification is enabled, then it should show the current message you are showing. But if the email notification is not enabled then it should show something like : `Spreadsheet download is successful!`

PR #1352, opened 2026-01-13
Fix: Visually distinguish disabled Clearing Request icon

### LINKED ISSUE #1340: CR Disabled/Enabled not visually different.
In the projects table(``` http://{domain}/projects ``` here), the "Create Clearing Request Icon" for projects for which CR can be created and for which CR is disabled are not visually different. User has no way of knowing unless they hover over the icon to see the disabled message.

Image

manhuu14:
### What this PR does
- Adds a clear visual distinction for the disabled "Create Clearing Request" icon
- Uses reduced opacity and disabled cursor to indicate non-interactive state
- Keeps tooltip explanation for better UX

### Issue
Fixes #1340

### DIFF
--- a/src/app/[locale]/projects/components/Projects.tsx
+++ b/src/app/[locale]/projects/components/Projects.tsx
@@ -272,12 +272,7 @@ function Project(): JSX.Element {
                                             <span className={'d-inline-block'}>
                                                 <BsCheck2Square
                                                     size={20}
-                                                    style={{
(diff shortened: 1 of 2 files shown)


manhuu14:
Hi @heliocastro Thanks for the suggestion. I’ve removed the inline styles and created a proper CSS class in the main stylesheet, which is now being used accordingly. 
Review it once Thank you !

PR #1353, opened 2026-01-13
fix(navbar): route SW360 logo to localized home page

### LINKED ISSUE #1338: Directs to wrong page on clicking icon in navbar
Clicking on SW360 icon in navabar should direct us to home page

Sandipmandal25:
The SW360 logo in the navbar was opening the wrong page.

## What changed
- Updated the SW360 navbar logo link to navigate to the localized home route (e.g. /en, /de).

This change makes the logo go to the home page for the current language.

Fixes #1338

### DIFF
--- a/src/components/sw360/Navbar/Navbar.tsx
+++ b/src/components/sw360/Navbar/Navbar.tsx
@@ -10,6 +10,7 @@
 
 'use client'
 
+import Link from 'next/link'
 import { useParams, useRouter, useSelectedLayoutSegment } from 'next/navigation'
 import { useSession } from 'next-auth/react'
 import { LocaleSwitcher, Logo, ProfileDropdown } from 'next-sw360'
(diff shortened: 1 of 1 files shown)


deo002:
It should redirect to the /home page

deo002:
It redirect to the same page as before

heliocastro:
Your code was not wrong, only @deo002  noticed another thing. The ancient original behavior is return to signing page, we already talked about change long time ago, but was in the middle of the development and still, this maybe can be configurable.

Sandipmandal25:
> Your code was not wrong, only @deo002 noticed another thing. The ancient original behavior is return to signing page, we already talked about change long time ago, but was in the middle of the development and still, this maybe can be configurable.

Thanks for the clarification @heliocastro !

  Just to confirm before I update the code should clicking the SW360 logo redirect to /${locale}/home instead of /${locale} (sign-in page)?

heliocastro:
Yes, the current target is home page, there's a caveat, though.
If you are not signed in, you should return to signig page ( like now ) or simply disable the link

PR #1356, opened 2026-01-13
fix: Pagination and sorting in Search Vendors Modal

### LINKED ISSUE #1349: Search Vendors Modal
## Description

Pagination and Sorting is not working in Search Vendors Modal

manhuu14:
## Description
Fixes pagination and sorting functionality in the Search Vendors Modal.

## Problem
The issue reported that pagination and sorting were not working in the Search Vendors Modal (#1349).

## Root Cause Analysis
After investigating the code, I found the root cause was a **state management issue** rather than just a simple bug:

1. The `useEffect` hook was missing critical dependencies (`currentPage` and `searchText`)
2. When sorting changed, it would use stale search text values
3. Pagination operations weren't maintaining the search context properly

## Solution Approach
While this is marked as a "good first issue," the fix required addressing the underlying architectural problem to ensure a robust solution:

### Changes Made:
- **Introduced `lastSearchedText` state**: Separates the user's input field (UI state) from the committed search query (data state)
- **Fixed useEffect dependencies**: Now properly depends on `lastSearchedText` to avoid stale closures
- **Ensured consistent behavior**: All pagination and sorting operations now use the same search context
- **Added proper state synchronization**: Search, sort, and pagination now work together correctly

###Fixes #1349

### DIFF
--- a/src/components/sw360/SearchVendorsModal/SearchVendorsModal.tsx
+++ b/src/components/sw360/SearchVendorsModal/SearchVendorsModal.tsx
@@ -104,7 +104,6 @@ function SearchVendorsModal({ chooseVendor }: Props): JSX.Element {
 
     useEffect(() => {
         if (session?.user?.access_token) {
-            setCurrentPage(0)
             void fetchVendors(0, lastSearchedText, sortField, sortDirection)
         }
     }, [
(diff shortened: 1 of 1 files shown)


heliocastro:
@manhuu14 Some of the inputs from copilot are valid. Please take a good look

manhuu14:
Hi @amritkv,
Conflicts resolved! The branch is now up-to-date with main and ready for review. 
Thanks!

PR #1366, opened 2026-01-14
fix: Fixed lucene search bug

### LINKED ISSUE #1350: Search Users Modal
Lucene Search query parameter is set as false even when the exact match is selected in the ``` Search Users ``` Modal.

sengmonkham:
#1350
# Description

This PR fixes the logic for the `luceneSearch` query parameter in the `SelectUsersDialog` component. Previously, selecting "Exact Match" would incorrectly set `luceneSearch` to `false` due to an inverted boolean logic (`!exactMatch`). This prevented users from performing exact match searches using Lucene.

The logic has been corrected to map `luceneSearch` directly to `exactMatch`.

Fixes #1350

## Type of change
Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/app/[locale]/projects/components/LinkProjects.tsx
+++ b/src/app/[locale]/projects/components/LinkProjects.tsx
@@ -323,7 +323,7 @@ export default function LinkProjects({
                             ? {}
                             : {
                                   name: searchValue,
-                                  luceneSearch: !exactMatch,
+                                  luceneSearch: exactMatch,
                               }),
                     }).map(([key, value]) => [
                         key,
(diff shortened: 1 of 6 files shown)


sengmonkham:
> Please sign your commits. DCO is required.

I have singed my commits, can you please check.

heliocastro:
@sengmonkham Can i ask why you have two signoffs with same email and different persons ?

sengmonkham:
Sorry about that. Those two are my own names, one is a pen name which I accidentally signed with the first time. Sorry for the confusion.

amritkv:
Hey @sengmonkham !
I have to close this PR as the linked issue (i.e. https://github.com/eclipse-sw360/sw360-frontend/issues/1350) itself created with some mis-understandings.
This is the expected behavior with CouchDB-nouveau. When the `Exact Match` is Checked, the `luceneSearch=false` else `luceneSearch=true`

PR #1370, opened 2026-01-14
fix: Use batch API for license clearing count to prevent 502 error

### LINKED ISSUE #1341: 502 for License Clearing Count Request
## Description

Request for license clearing count in the projects table ( ``` https://{domain}/projects ```  here) throws 502 frequently if we increase the number of page entries. 

Image

## How to reproduce?

- Change the number of page entries to 100
- Change table pages

taanvi2205:
Fixes #1341 
Previously each row in the projects table made an individual API call to fetch license clearing count is causing 502 errors when page size was increased to 100. Now uses a single batch POST request to fetch all clearing counts at once.

### DIFF
--- a/src/app/[locale]/projects/components/Projects.tsx
+++ b/src/app/[locale]/projects/components/Projects.tsx
@@ -33,7 +33,6 @@ import {
 } from '@/object-types'
 import MessageService from '@/services/message.service'
 import { ApiUtils, CommonUtils } from '@/utils'
-import { SW360_API_URL } from '@/utils/env'
 import ImportSBOMMetadata from '../../../../object-types/cyclonedx/ImportSBOMMetadata'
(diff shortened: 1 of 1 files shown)


src/app/[locale]/projects/components/Projects.tsx
```diff
@@ -390,6 +404,58 @@ function Project(): JSX.Element {
...
+
+                const response = await fetch(`${SW360_API_URL}/resource/api/projects/licenseClearingCount`, {
+                    method: 'POST',
+                    headers: {
+                        'Content-Type': 'application/json',
+                        Authorization: session.user.access_token,
+                    },
+                    body: JSON.stringify(projectIds),
+                    signal,
+                })
```
Comment (heliocastro) [inline src/app/[locale]/projects/components/Projects.tsx]: Please use fetch from APIUtils functions

Issue #1382, opened 2026-01-16
fix(CR): Bring back `tag` field at clearing request table


amritkv:
Bring back `tag` field in clearing request table.

New backend PR which send tag field : https://github.com/eclipse-sw360/sw360/pull/3593
Linked recent UI PR :  https://github.com/eclipse-sw360/sw360-frontend/pull/1314

PR #1383, opened 2026-01-16
fix(CR): Add tag field in CR tables

akshitjoshii:
Closes: #1382

Added `tag` field back in the open and closed clearing requests table.

Screenshot from 2026-01-16 15-09-23

### DIFF
--- a/src/app/[locale]/requests/components/ClearingRequest.tsx
+++ b/src/app/[locale]/requests/components/ClearingRequest.tsx
@@ -77,7 +77,17 @@ function ClearingRequestComponent({ requestType }: { requestType: RequestType })
                 header: t('BA-BL/Group'),
                 cell: ({ row }) => <>{row.original.projectBU ?? t('Not Available')}</>,
             },
-
+            {
+                id: 'tag',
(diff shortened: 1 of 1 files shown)


src/app/[locale]/requests/components/ClearingRequest.tsx
```diff
@@ -77,7 +77,17 @@ function ClearingRequestComponent({ requestType }: { requestType: RequestType })
...
-
+            {
+                id: 'tag',
+                header: t('Tag'),
+                cell: ({ row }) => {
+                    if (!Object.hasOwn(row.original, 'projectId')) {
+                        return <>{t('Project Deleted')}</>
+                    } else {
+                        return <>{row.original._embedded?.['sw360:projectDTOs']?.[0]?.tag ?? ''}</>
+                    }
```
Comment (amritkv) [inline src/app/[locale]/requests/components/ClearingRequest.tsx]: Hey @akshitjoshii !
If the project gets deleted or if `tag` is not present, then the `tag` field should be: `Not Available` instead of `Project Deleted`.

May be something like the `BA-BL/Group` column

amritkv (approved):
LGTM !

PR #1393, opened 2026-01-19
fix(UI): Vendor field shows empty on project edit page even when vendor data exists.

nikkuma7:
## Steps to Test
1. Create a project with vendor information
2. Save the project
3. Navigate to edit the project (e.g., `/projects/edit/{id}?tab=summary`)
4. Observe that the vendor field is not empty. (eg . firstname)

image


deo002 (changes requested):
Changes look good. Working as expected. Just reformatting is required.

deo002 (approved):
LGTM!

Issue #1384, opened 2026-01-16
fix(Adv Search): Add `luceneSearch` param to advanced search


amritkv:
Add `luceneSearch` field in the API call while using Advanced Search in different pages.

Make sure to put `luceneSearch=true` if `exactMatch=false` and vice-versa.

PR #1397, opened 2026-01-20
fix(AdvSearch): Fixed luceneSearch param in adv search

amritkv:
Fixed luceneSearch param in advanced search:
-  If `Exact Match` is checked then `luceneSearch=false`
- Else if `Exact Match` is un-checked then `luceneSearch=true`

Closes : #1384


heliocastro (changes requested):
@amritkv As you are touching the code, please move inline styles

amritkv:
> @amritkv As you are touching the code, please move inline styles

Hey @heliocastro, applied the requested changes.

PR #1425, opened 2026-01-28
Fix license table empty shortName and broken detail link

### LINKED ISSUE #1318: License table has empty short name and no direct link for the license view

Sandipmandal25:
## Summary                                                                                                                            
  - License table showed blank text when shortName was empty or missing                                                                 
  - Clicking a license row did not navigate to the detail page correctly because the link was built from shortName instead of the actual license ID                                                                                                                           
                                                                                                                                        
  ## Changes                                                                                                                            
  - Extract license ID from `_links.self.href` instead of relying on shortName for the detail link                                      
  - Added `accessorKey` to the shortName column                                                                                         
  - Added fallback display so the cell never shows blank text                                                                           
  - Used `encodeURIComponent` on the link parameter                                                                                     
                                                                                                                                        
  Closes #1318

### DIFF
--- a/src/app/[locale]/licenses/components/LicensePage.tsx
+++ b/src/app/[locale]/licenses/components/LicensePage.tsx
@@ -81,15 +81,20 @@ function LicensePage(): ReactNode {
         () => [
             {
                 id: 'shortName',
+                accessorKey: 'shortName',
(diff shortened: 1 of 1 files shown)


heliocastro (approved):
Good to go

PR #1460, opened 2026-02-12
Revert "refactor(styles): Unify all styles in globals"

deo002:
This reverts commit 95a122b6c83807be0301aa984785bc06c02ed33e. It breaks css in several components. It is easier to revert this commit and do it from scratch again.


amritkv (approved):
LGTM !

PR #1493, opened 2026-02-24
fix(requests): Show error message on edit Clearing Request failure

### LINKED ISSUE #1492: fix(requests): Fix error message at edit Clearing Requests
Currently at edit CR page, if we change say only the CR state for exmaple `OPEN` to `CLOSED` and did not select the agreed clearing date, and try to update it, the UI shows no change or no error message. 
But actually under the hood its a 400 error saying `"Invalid agreed clearing date requested"`.

In such case we need to show proper banner message with error message in the UI.

Sandipmandal25:
## Summary
  - Show backend error message to user when updating a Clearing Request fails
  - Previously, non-OK responses (e.g. 400 "Invalid agreed clearing date requested") were silently caught with no UI feedback
  - Added `else` branch to display `MessageService.error(data.message)`, same pattern as EditComponent and EditRelease

  Fixes #1492


amritkv (approved):
LGTM !

PR #1536, opened 2026-03-19
fix(hooks): Guard useLocalStorage against invalid JSON

### LINKED ISSUE #1535: useLocalStorage crashes when localStorage contains malformed JSON
## Problem
`useLocalStorage` directly calls `JSON.parse` on stored values.  
If the value is malformed JSON, it throws and can break page rendering.

## Affected File
- `src/hooks/useLocalStorage.ts`

## Steps To Reproduce
1. In browser devtools, run:
   `localStorage.setItem('uiConfig', '{bad-json')`
2. Open a page that reads this key via `useLocalStorage`.
3. Observe a runtime `SyntaxError` from `JSON.parse`.

ran into a bug in src/hooks/useLocalStorage.ts. If localStorage contains invalid JSON (which can happen if a user messes with it or if a previous version of the app saved data differently), the whole hook crashes because it doesn't handle JSON.parse errors.

YianZhao:
## Summary
This PR hardens `useLocalStorage` against malformed JSON values stored in `localStorage`.

## Related Issue
Fixes #1535 

## Changes
- Added safe parser helper inside `useEffect`
- Wrapped `JSON.parse` in `try/catch`
- Fallback to `initialValue` on parse failure
- Remove invalid storage key to prevent repeated failures


GMishx (approved):
LGTM

PR #1560, opened 2026-03-29
fix(duplicate): projectManager not populated in projectPayloadData on duplicate project page

### LINKED ISSUE #1559: fix(duplicate): projectManager not populated in projectPayloadData on duplicate project page
## Bug Description

In `src/app/[locale]/projects/duplicate/[id]/components/DuplicateProject.tsx`, 
the `projectManager` field is correctly fetched and stored in local state, 
but it was never assigned in `projectPayloadData` — the object sent to the API.

`projectOwner` and `leadArchitect` were both assigned but `projectManager` 
was missing entirely.

## Impact

When duplicating a project, the Project Manager is silently dropped — 
the duplicated project always has an empty Project Manager field regardless 
of what the source project had.

## Fix

Add the missing line to `projectPayloadData`:
```typescript
projectManager: project._embedded?.projectManager?.email ?? '',
```

sujalkalauni:
## Summary

In `DuplicateProject.tsx`, the `projectManager` field was correctly fetched 
from the API and stored in local state, but was never assigned in 
`projectPayloadData` — the object actually sent to create the duplicate project.

This meant every duplicated project silently lost its Project Manager.

## Fix

Added the missing assignment alongside the existing `projectOwner` and 
`leadArchitect` fields:
```typescript
projectManager: project._embedded?.projectManager?.email ?? '',
```

## Related

Closes #1559


sujalkalauni:
Thanks for the review @amritkv. I can see both projectResponsible (top-level string) and projectManager (User object in _embedded) in the Project type. I followed the same pattern as leadArchitect and projectOwner which also use _embedded.
Could you help me understand which one maps to the Project Manager field in the UI? Should I use project.projectResponsible instead?

amritkv:
Hey @sujalkalauni !
The test is still failing. The `Project Manager` filed isn't showing up with existing values.

sujalkalauni:
Hey @amritkv, could you share which test is failing so I can reproduce and fix it?

GMishx (approved):

PR #1577, opened 2026-04-03
fix(header): Restrict csp headers

rudra-superrr:
Summary
Harden the Content-Security-Policy header in next.config.ts as recommended by CRA security assessment.

Problem
The CSP header was using insecure keywords unsafe-eval in the script-src directive unconditionally, and was missing restrictive directives like object-src.

Note:  'unsafe-inline' is retained in script-src as inline scripts are currently used in the codebase. The proper fix is to implement a nonce-based CSP- generating a per-request nonce in the middleware, propagating it via the root layout, and refactoring the inline script usages to use safe rendering patterns.

Testing
Start the server
Check headers via curl
curl -sI http://localhost:3000/ | grep -i "content-security-policy"


GMishx:
Done in #1664, #1640 and #1639

PR #1589, opened 2026-04-06
fix(Vul): Fixed release detail url at vul tracking tab

amritkv:
Fixed release detail url at `Vulnerability Tracking Status` tab under project detail page.


deo002 (approved):
LGTM!

PR #1592, opened 2026-04-06
fix(SearchUsersDialog): Improve initial load handling and caching for user search

aaryan359:
## Summary
Fix search dialogs so users do not need to click **Search** with an empty value to load results.

### What Changed
- Load initial results automatically when dialog opens.
- Remove undefined-search bootstrap behavior.
- Improve user search handling with in-memory caching to reduce repeated API calls.

### Impact
- Dialogs open with data immediately.
- Fewer unnecessary backend calls.
- Better and more consistent UX across role/vendor selectors.

## Before
[Screencast from 2026-04-06 23-39-04.webm](https://github.com/user-attachments/assets/c63b25a3-7db5-4816-85b0-b2c11051b89f)

## After
[Screencast from 2026-04-07 00-13-38.webm](https://github.com/user-attachments/assets/8f8b224d-dfc7-4261-be06-538baa72f3aa)

### DIFF
--- a/src/components/sw360/SearchUsersDialog/SelectUsersDialog.tsx
+++ b/src/components/sw360/SearchUsersDialog/SelectUsersDialog.tsx
@@ -17,7 +17,7 @@ import Link from 'next/link'
 import { signOut, useSession } from 'next-auth/react'
 import { useTranslations } from 'next-intl'
 import { PageSizeSelector, SW360Table, TableFooter } from 'next-sw360'
-import { type JSX, useEffect, useMemo, useState } from 'react'
+import { type JSX, useEffect, useMemo, useRef, useState } from 'react'
 import { Button, Col, Form, Modal, Row, Spinner } from 'react-bootstrap'
 import { BsInfoCircle } from 'react-icons/bs'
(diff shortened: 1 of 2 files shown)


GMishx:
Hey @aaryan359 , thank you for the PR. But this is an undesirable behavior. We are putting unnecessary load on the backend without even understanding user's intent.

aaryan359:
@GMishx, This pr is not putting load on the backend. The main issue is that when we open any dialog box, like for entering a user or a project that we have already present in the database.
But this will not show to the user until he explicitly presses the search button, until then, he will think that nothing is present in the DB. 

But it must load initially and then store in the cache, so no load on the backend.
and this is exact this pr is doing.

GMishx:
@aaryan359 that's a nice thinking. But you are imagining 1 client 1 server. But in reality, there are 10,000 clients 1 server. Now imagine each client opening these modals and just loading these data only to put in actual search term and discarding what is there, which is actually going to hit backend anyways. Caching won't help as this has to be near real time (how you'd explain to the user the new data they just created is not visible?).

PR #1594, opened 2026-04-07
fix(modals): Reset pagination on search in modals

deo002:
Reset pagination on search in modals. It might be possible that the user is on page 10 when he decides to perform a search. The search result might have only 2 pages, leading to errors in application. Thus, the pagination must be reset.


amritkv (approved):
LGTM !

PR #1613, opened 2026-04-10
fix(components): handle empty response in release overview

bibhuti230185:
## Summary
Fixes JSON parse error when the backend returns an empty response body in the Release Overview component.

## Problem
When viewing a component with no releases, the API returns an empty response body (HTTP 200 with no content). The previous implementation attempted to call `response.json()` directly, causing a `SyntaxError: Unexpected end of JSON input`.

image

## Solution
- Check response text before parsing JSON
- Use optional chaining (`?.`) for safer property access on `_embedded`
- Return empty array when response body is empty

## Changes
- [ReleaseOverview.tsx](src/app/[locale]/components/detail/[id]/components/ReleaseOverview.tsx#L226-L233)

## Suggested Reviewer
@amritkv  @deo002 

## Testing
1. Navigate to a component detail page with no releases
2. Verify no console errors appear
3. Release overview table shows "No releases found" (or similar empty state)

### DIFF
--- a/src/app/[locale]/components/detail/[id]/components/ReleaseOverview.tsx
+++ b/src/app/[locale]/components/detail/[id]/components/ReleaseOverview.tsx
@@ -223,12 +223,14 @@ const ReleaseOverview = ({ componentId, calledFromModerationRequestDetail }: Pro
                     })
                 }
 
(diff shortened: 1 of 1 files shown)


bibhuti230185:
image

amritkv (approved):
LGTM !

PR #1615, opened 2026-04-10
fix(ExportSpreadsheet): Fixed export spreadsheet in project with search results

Farooq-Fateh-Aftab:
> Please provide a summary of your changes here.
> * Which issue is this pull request belonging to and how is it solving it? (*https://github.com/eclipse-sw360/sw360-frontend/issues/1614*)
> * Did you add or update any new dependencies that are required for your change?

Issue: https://github.com/eclipse-sw360/sw360-frontend/issues/1614

### How To Test?
    1. Navigate to Project Tab.
    2. Search with Advance filter.
    3. Export the spreadsheet with both option, excel will have the same result as shown in UI after search.


src/app/[locale]/projects/components/Projects.tsx
```diff
@@ -575,7 +575,12 @@ function Project(): JSX.Element {
             if (CommonUtils.isNullOrUndefined(session)) return signOut()
 
             const mailEnabled = mailRequestForProjectReport === 'true'
-            const url = withLinkedRelease ? 'reports?module=PROJECTS&withLinkedRelease=true' : 'reports?module=PROJECTS'
+            const searchParams = Object.fromEntries(params.entries())
+            const searchParamsString = new URLSearchParams(searchParams).toString()
```
Comment (amritkv) [inline src/app/[locale]/projects/components/Projects.tsx]: Hey @Farooq-Fateh-Aftab !
We can combine these 2 lines into one liner.
It could be like `const searchParamsString = params.toString()` instead of doing `params → object → string`.
The current approach is suitable when we need to modify the param, but here it's not the case.

amritkv (changes requested):

amritkv (approved):
LGTM !

PR #1637, opened 2026-04-15
fix(components): handle empty response in edit component page

bibhuti230185:
## Summary

Fixes JSON parse error when the backend returns an empty response body in the Edit Component page for releases and attachments.

Related to: #1613

## Problem

When editing a component with no releases or attachments, the API returns an empty response body (HTTP 200 with no content). The previous implementation attempted to call `response.json()` directly, causing a `SyntaxError: Unexpected end of JSON input`.

This is the same issue that was fixed in PR #1613 for the component detail page, but the Edit Component page (`EditComponent.tsx` and `Releases.tsx`) was not addressed.

## Solution

- Check response text before parsing JSON
- Use optional chaining (`?.`) for safer property access on `_embedded`
- Return empty array when response body is empty

## Changes

- `src/app/[locale]/components/edit/[id]/components/Releases.tsx`
- `src/app/[locale]/components/edit/[id]/components/EditComponent.tsx`

## Suggested Reviewer

@amritkv @deo002

## Testing

1. Navigate to Edit Component page for a component with no releases
2. Verify no console errors appear
3. Releases table shows empty state correctly

Local Test screenshots:
image

image

### DIFF
--- a/src/app/[locale]/components/edit/[id]/components/EditComponent.tsx
+++ b/src/app/[locale]/components/edit/[id]/components/EditComponent.tsx
@@ -124,10 +124,15 @@ const EditComponent = ({ componentId }: Props): ReactNode => {
                 } else if (response.status !== StatusCodes.OK) {
                     return notFound()
                 }
-                const dataAttachments: EmbeddedAttachments = (await response.json()) as EmbeddedAttachments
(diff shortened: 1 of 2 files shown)


amritkv (approved):
LGTM !

PR #1638, opened 2026-04-16
fix(filter): Fix filter in projects attachment usage

deo002:
## Fixes:
- The filter 'Without CLI Usage Set' was not working as project path was not being added to the license info string while searching in save usages payload.
- AttachmentUsages payload was sent as a matrix from the backend which is no longer the case. It is sent as an array now.


amritkv (approved):
LGTM !

PR #1679, opened 2026-05-06
Fix loading of releases and update of obligation comment

deo002:


GMishx (approved):
Tested, working as expected.

PR #1681, opened 2026-05-07
fix(obligations): Add read-only all obligations tab in project details

deo002:
Frontend PR for https://github.com/eclipse-sw360/sw360/pull/3997


amritkv (approved):
LGTM !

PR #1696, opened 2026-05-11
feat(advancesearch): Allow Info button for labels

GMishx:
Allow putting a new _(i)_ info button next to field of Advance Search component.


GMishx:
The feature does not, but the messages require https://github.com/eclipse-sw360/sw360/pull/4153 to work actually.

amritkv (approved):
LGTM !

PR #1697, opened 2026-05-12
fix(changelog): Remove inner html

### LINKED ISSUE #1677: Remove innerHTML from code
There are anti patterns of using `innerHTML` in `src/components/ChangeLog/CreateChangeCard.ts` which must be removed and use React best practices.

deo002:
Closes #1677 

The old component used innerHtml for rendering components. The old and new values were converted into strings and diff was calculated. The difference was hightlighted by appending span components to the diffed words via string manipulation. React treats the span-manipulated strings as strings and not html tags when not set via innerHtml tag.

So, I had to create a new component which didn't use string manipulation to highlight diffs.


amritkv:
Hey @deo002 !
Checked the code, it LGTM !

But we're missing one feature which was added in this PR : https://github.com/eclipse-sw360/sw360-frontend/pull/1680

deo002:
Oops! Thanks for pointing out!

deo002:
Done! @amritkv

amritkv (approved):
Changes LGTM !

PR #1698, opened 2026-05-13
fix(package): Package deletion infinite loading issue

Suhas2109:
After deleting a package from the edit page, the browser would show an infinite loading spinner instead of displaying a success message and redirecting properly.

Before Changes:
image

After Changes:
image


amritkv (approved):
LGTM !

PR #1716, opened 2026-05-16
fix(frontend): Remove Trusted Types CSP directive

GMishx:
## Summary
Removes the `require-trusted-types-for 'script'` directive from the CSP header configuration. Testing revealed that Next.js (Turbopack-based runtime) currently does not support Trusted Types enforcement, leading to `HTMLScriptElement.src setter` violations that block chunk loading during development.

## Root Cause
When `require-trusted-types-for 'script'` is enforced, any script sink that assigns a plain string to `script.src` is blocked. Next.js runtime chunks (specifically Turbopack runtime code in `/_next/static/chunks/`) assign script URLs as plain strings during dynamic chunk loading, triggering CSP violations.

CSP Violation Details:
- effectiveDirective: "require-trusted-types-for"
- violatedDirective: "require-trusted-types-for"
- blockedURI: "trusted-types-sink"
- sample: "HTMLScriptElement src|/_next/static/chunks/..."

## Why We Can't Enable It Now
1. Next.js Framework Limitation: Turbopack and the React runtime do not implement Trusted Types APIs for script sink assignment.
2. All Script Paths Affected: The violation occurs on core runtime paths, not third-party dependencies or custom code.
3. No Per-Path Workaround: Unlike other CSP directives, TT requires either framework-level support (outside our control) or custom Trusted Types policies that significantly weaken the security benefit.

## What Was Tested
- Enabled TT in Report-Only mode to assess violations without breaking the app
- Tested production build (pnpm build + .next/standalone/server.js)
- Confirmed violations originate from Next.js runtime (not app code)
- Verified no regressions in other CSP directives (HSTS, X-Frame-Options, etc.)

## Recommendation
- Current: Keep existing CSP hardening without TT enforcement
- Future: Revisit enforcement once Next.js/Turbopack implements Trusted Types support

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -22,7 +22,6 @@ const csp = `
   connect-src 'self' https://www.gravatar.com${isDev ? ' http://localhost:*' : ''};
   object-src 'none';
   frame-ancestors 'self';
-  ${isDev ? '' : "require-trusted-types-for 'script'"};
 `
     .replace(/\s{2,}/g, ' ')
     .trim()


deo002 (approved):
LGTM!

PR #1718, opened 2026-05-16
chore(docker): Add jwk-set-uri option

GMishx:
Sync docker config updates from https://github.com/eclipse-sw360/sw360/pull/4170


deo002 (approved):
LGTM!

PR #1727, opened 2026-05-17
fix(ci): Restrict contents write permission to job

GMishx:
Attempt to fix https://github.com/eclipse-sw360/sw360-frontend/security/code-scanning/57

This restricts the `content: write` to only the job which needs it, not the entire workflow.


amritkv (approved):
LGTM !

PR #1730, opened 2026-05-19
feat(search): Search by lucene ranking based on score

rudra-superrr:
Problem Description

When users perform a search (advanced search or modal search) in the SW360 frontend, the results returned from the backend are ranked by **Lucene relevance score**. However, the frontend was not passing the appropriate `sort` parameter to the REST API during search requests. This caused search results to appear in an arbitrary or default column order (e.g., alphabetical by name) instead of being sorted by relevance.

Solution
Conditionally set `sort: 'score,asc'` whenever a search is active, and reset to the default sort (empty string or column-based) when no search is active.

Testing

1. **Search relevance**: Perform a search on any list page → results should appear with the most relevant match first.
2. **No search**: Clear the search → results should return to default ordering (alphabetical or server default).
3. **Column sort after search**: Click a column header after searching → results should sort by that column (overriding score sort).

Note: This PR requires the backend PR(https://github.com/eclipse-sw360/sw360/pull/4200) to be deployed as well for testing.

### DIFF
--- a/src/app/[locale]/admin/bulkreleaseedit/components/ReleaseEditPage.tsx
+++ b/src/app/[locale]/admin/bulkreleaseedit/components/ReleaseEditPage.tsx
@@ -449,7 +449,7 @@ export default function BulkReleaseEdit(): JSX.Element {
         setPageableQueryParam({
             page: 0,
             page_entries: 10,
-            sort: '',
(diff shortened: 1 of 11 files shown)


amritkv (approved):
LGTM !

PR #1731, opened 2026-05-19
fix(ci): Migrate pnpm config

GMishx:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

## What changed

This PR removes deprecated pnpm override configuration from `package.json` and aligns lockfile metadata with current pnpm behavior.

- removed legacy `pnpm.overrides` configuration from `package.json`
- kept `ignoredBuiltDependencies` as explicit install policy
- refreshed `pnpm-lock.yaml` so frozen installs match active config

## Why this is safe

The previous overrides were not materially changing resolution in the current dependency graph; they were acting mostly as policy guardrails from an earlier migration period.

### Evidence from dependency inspection

- `@eslint/plugin-kit` override: not present in active resolved graph
- `biome` override: not present as an active transitive package entry needing override
- `@types/react` / `@types/react-dom` overrides: redundant with direct `devDependencies` pins
- `lodash` override: functioned as a version floor policy, not required to make install succeed

After removing overrides and regenerating lock metadata, installs still resolve and complete under frozen lockfile mode.

## CI issue addressed

CI was failing with:

`ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`

because pnpm v10 no longer reads `pnpm.overrides` from `package.json`, while lockfile metadata still contained override state.

## How to test

```bash
pnpm install --no-frozen-lockfile
pnpm install --frozen-lockfile
```

**Expected:** frozen install succeeds without lockfile config mismatch.

### Impact
- No user-facing behavior change
- No runtime feature change
- CI/package-management config cleanup for pnpm v10 compatibility

### Checklist
- [x] Lockfile updated after config cleanup
- [x] Frozen install validated locally
- [x] AI-assistance disclosure added if required by project policy

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,19 +94,6 @@
       ]
     }
   },
-  "pnpm": {
-    "overrides": {
-      "@types/react": "19.2.14",
-      "@types/react-dom": "19.2.3",
-      "@eslint/plugin-kit": "^0.3.3",
-      "biome": "-",
-      "lodash": ">=4.18.0"
-    },
-    "ignoredBuiltDependencies": [
-      "cypress",
-      "sharp"
-    ]
-  },
   "prettier": {
     "semi": false,
     "tabWidth": 4,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,13 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  '@types/react': 19.2.14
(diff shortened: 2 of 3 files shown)


amritkv (approved):
LGTM !

PR #1733, opened 2026-05-20
fix(projects): Handle Unicode in report download filename and correct file extension

bibhuti230185:
## Problem

When generating a license-info report for a project whose name contains non-ASCII characters (e.g. `™`, umlauts), two issues occurred:

1. **Wrong file extension**: The downloaded file always had `.zip` extension regardless of the selected output format (DOCX, XHTML, TEXT).
2. **Unicode filename loss**: The backend `Content-Disposition` header was stripped by Tomcat when project names contained non-ISO-8859-1 characters, causing the browser to fall back to the generic frontend filename.

### `src/services/download.service.ts`
- Added `parseFilenameFromContentDisposition()` utility that extracts the filename from the response header
- Supports RFC 5987 `filename*=UTF-8''...` (preferred) and plain `filename="..."` (fallback)
- Download now uses the server-provided filename when available, falling back to the caller-supplied name

### `src/app/[locale]/projects/generateLicenseInfo/[id]/components/DownloadLicenseInfoModal.tsx`
- File extension is now derived from the selected `generatorClassName`:
  - `DocxGenerator` → `.docx`
  - `XhtmlGenerator` → `.html`
  - `TextGenerator` → `.txt`
  - fallback → `.zip`

## Backend PR
https://github.com/eclipse-sw360/sw360/pull/4202
## Testing

- Created a project with Unicode name (`Gridscale X™ Protection Data Manager`)
- Generated license-info report with DOCX format
- Verified file downloads with correct `.docx` extension
- Verified RFC 5987 header parsing works when backend sends encoded filename

https://github.com/user-attachments/assets/3c75e82c-8e7f-407a-865c-c666fd66ee4e

### DIFF
--- a/src/app/[locale]/projects/generateLicenseInfo/[id]/components/DownloadLicenseInfoModal.tsx
+++ b/src/app/[locale]/projects/generateLicenseInfo/[id]/components/DownloadLicenseInfoModal.tsx
@@ -137,10 +137,16 @@ export default function DownloadLicenseInfoModal({
                 variant: 'DISCLOSURE',
                 selectedRelRelationship,
             })
+            const extensionMap: Record<string, string> = {
+                DocxGenerator: 'docx',
+                XhtmlGenerator: 'html',
+                TextGenerator: 'txt',
+            }
+            const ext = extensionMap[generatorClassName] ?? 'zip'
             const downloadStatus = await DownloadService.download(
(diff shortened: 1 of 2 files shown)


amritkv (approved):
Tested and its working as expected.

PR #1756, opened 2026-06-01
fix(user): Add externalid upon create and edit of user

### LINKED ISSUE #1755: External id is missing in create and update user
External id is missing in create and update user

Farooq-Fateh-Aftab:
# fix(user): Add externalid upon create and edit of user

## Summary of Changes

The `externalid` (Global Identifier) field was not being populated when editing a user and was not being sent correctly when creating a user due to a wrong `name` attribute in the form input.

**Changes:**
- **`UserEditForm.tsx`**: Fixed the Global Identifier input — changed `name='globalIdentifier'` → `name='externalid'` and `type='globalIdentifier'` → `type='text'`, so `handleChange` writes to the correct `UserPayload.externalid` field
- **`edit/[id]/page.tsx`**: Added `externalid: user.externalid` when mapping the API response into `userPayload`, so the field is pre-populated on the edit form
- **`add/page.tsx`**: Added `externalid: ''` to the initial state so the field is tracked from the start and included in the POST request

---

## Issue

> Global Identifier (externalid) is not fetched/displayed on the edit user page and is not saved when creating or updating a user.

closes: https://github.com/eclipse-sw360/sw360-frontend/issues/1755

---

## Suggest Reviewer

> @GMishx, @amritkv 

---

## How To Test?

1. Navigate to **Admin → Users → Add User**
   - Fill in all fields including **Global Identifier**
   - Submit → verify the new user is created with the correct `externalid` via the user detail page

2. Navigate to **Admin → Users → Edit** an existing user
   - Verify the **Global Identifier** field is pre-populated from the API
   - Change the value and save → verify the updated `externalid` is reflected on the user detail page

---

### DIFF
--- a/src/app/[locale]/admin/users/add/page.tsx
+++ b/src/app/[locale]/admin/users/add/page.tsx
@@ -29,6 +29,7 @@ export default function CreateUser(): JSX.Element {
         department: '',
         fullName: '',
         password: '',
+        externalid: '',
         secondaryDepartmentsAndRoles: {},
     })
     const router = useRouter()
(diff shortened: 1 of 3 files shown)


amritkv (approved):
LGTM !

PR #1758, opened 2026-06-01
fix(licenseClearing): Handle no releases

GMishx:
Handle `/licenseClearing` response where there are no embedded releases. Also update the object type to indicate the value can be undefined.


amritkv (approved):
LGTM !

PR #1766, opened 2026-06-03
fix(sort): Add sorting to license clearing page

deo002:
- First list all Releases sorted by name in ascending order.
- Then list all Projects sorted by name in ascending order.
- When a Project is expanded, the same order (1, 2) should be followed.


amritkv (approved):
LGTM !

PR #1773, opened 2026-06-05
fix(redirect): Redirect user to path after auth

GMishx:
If user opens a direct link for the first time, they are redirected to home page with `callbackUrl=` parameter. Once they come back authenticated, this value is ignored.

This change makes sure the user is redirected back to localized path once authenticated.


amritkv (approved):
Tested and working fine.

PR #1776, opened 2026-06-05
fix(color): Applied soften colors globally

### LINKED ISSUE #1775: fix(color): Apply soften color for license clearing PS/CS capsules
Apply soften color for license clearing PS/CS capsules. The current success, warning, danger colors are very bright and not softened color as it used to be in existing UI.

amritkv:
### What
Override Bootstrap's default `.bg-success`, `.bg-danger`, `.bg-warning`, and `.bg-secondary` background colors with softer, solid SW360 values for the project/release state badges (the `PS`/`CS` capsules) in `globals.css`

### Why
The previous custom overrides weren't applying — the badges rendered in Bootstrap's harsh default colors instead of the intended softer palette.

### Root cause
The custom `.bg-success` rule and Bootstrap's `.bg-success` both used `!important` with identical specificity `(0,1,0)`. With importance and specificity tied, the cascade falls back to source order, and Bootstrap's stylesheet loads last in the final `<head>`, so Bootstrap won.

### How
Scoped each override to `.badge.bg-*`, raising specificity to `(0,2,0)` so it beats Bootstrap's `(0,1,0)` on specificity directly — making the fix independent of stylesheet import order. 

Closes : #1775

### DIFF
--- a/src/styles/globals.css
+++ b/src/styles/globals.css
@@ -567,6 +567,22 @@ body {
     background-color: #ffd350 !important;
 }
 
+.badge.bg-success {
+    background-color: #69c17d !important;
+}
+
+.badge.bg-danger {
+    background-color: #e6717c !important;
+}
+
+.badge.bg-secondary {
+    background-color: #dee2e6 !important;
+}
+
+.badge.bg-warning {
+    background-color: #ffd350 !important;
+}
+
 .spinner {
     color: #f7941e;
 }
--- a/src/app/[locale]/projects/components/Projects.tsx
+++ b/src/app/[locale]/projects/components/Projects.tsx
@@ -1055,7 +1055,7 @@ function Project(): JSX.Element {
                 },
                 {
                     key: 'PHASE_OUT',
-                    text: t('PhaseOut'),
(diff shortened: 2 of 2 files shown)


src/styles/globals.css
```diff
@@ -551,19 +551,19 @@ body {
     border-bottom-left-radius: 0;
 }
 
-.bg-success {
-    background-color: #68c17c !important;
+.badge.bg-success {
```
GMishx, line 570: Why disturb the existing `bg-success`?

amritkv: @GMishx, this is because it is the common class for multiple places where pill like badges are used. And everywhere bootstrap is overriding our custom soft touch colors with its very sharp and bright default colors.

GMishx: My point being, this would remove the override we have for `.bg-success` class. Why not create another override of `.badge.bg-success` along with `.bg-success`?

amritkv: Done !

GMishx (approved):
LGTM

PR #1787, opened 2026-06-11
fix(frontend): Enforce valid use client directives

GMishx:
## Summary

This PR fixes frontend components affected by invalid or missing `'use client'` directives and adds guardrails so the same issue is caught before merge.

### Changes
- fix invalid directive typos such as `'use-client'` / `'used-client'`
- add missing `'use client'` to client-only components
- add a Husky pre-commit check for client directives
- run the same client-directive check in CI build workflow

### Client directive check

This PR adds a small check script to prevent invalid or missing client component directives.

The script does the following:

- scans staged frontend source files under `src/`
- fails if it finds invalid directive typos like:
  - `'use-client'`
  - `'used-client'`
- checks `.tsx` files in `src/app/` and `src/components/`
- if a file uses obvious client-only patterns, it requires `'use client'` as the first statement

Examples of client-only patterns the script looks for:

- `next-auth/react`
- `ApiUtils`
- browser globals such as `window`, `document`, or `location`
- React client hooks such as `useState`, `useEffect`, `useMemo`, and `useCallback`

This helps enforce the current frontend direction of client-side rendering only and prevents broken directive spellings from reaching CI or deployment.

## Why

Recent stage deployments exposed frontend requests being executed from the wrong runtime boundary. Since the current direction is client-side rendering only, client-only components should explicitly declare `'use client'`, and invalid spellings should fail fast in local checks and CI.

## Scope

This change is limited to the `sw360-frontend` repository.

## How to test

```bash
pnpm check:client-directives
pnpm build
```

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -56,6 +56,11 @@ jobs:
         run: |
           pnpm i
 
+      - name: Check client directives
+        shell: bash
+        run: |
+          pnpm check:client-directives
+
       - name: Run linting
         shell: bash
         run: |
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,2 @@
+pnpm check:client-directives
 pnpx lint-staged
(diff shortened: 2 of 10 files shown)


deo002 (approved):
LGTM!

PR #1774, opened 2026-06-05
fix(session): Centralize signout and tokens

GMishx:
## Summary

This PR contains 2 focused commits to stabilize auth/session behavior and reduce auth-related code duplication.

### 1) Centralized signout/session-expiry handling
- Added `SessionStatusHandler` at app provider level (`src/app/provider.tsx`).
- Introduced `dispatchSessionExpiredEvent` flow (`src/utils/sessionExpiry.utils.ts`) and routed session-expiry handling through one place.
- Result: signout/redirect behavior is centralized and consistent across pages/components.
- The `SessionProvider` can now be configured to have a session refresh interval with `SW360_SESSION_REFETCH_INTERVAL_SECONDS` with a fallback of 5 minutes.

### 2) API client split for authenticated vs unauthenticated calls
- Added:
  - `AuthenticatedApiUtils` (`src/utils/api/authenticatedApi.util.ts`)
  - `UnauthenticatedApiUtils` (core transport in `src/utils/api/api.util.ts`)
- Kept compatibility alias:
  - `ApiUtils` in `src/utils/index.ts` to minimize broad import churn in this PR.
- Added docs in `src/utils/index.ts` clarifying intended usage.

## Why this change

- Avoid scattered, duplicated session-expiry/signout logic in many components.
- Make API call intent explicit:
  - authenticated endpoints should trigger session-expiry flow on `401`,
  - unauthenticated endpoints (e.g. version/footer-style) should not force signout behavior.
- Keep this PR low-risk by avoiding mass import migration.

## Important note on `ApiUtils` alias

`ApiUtils` alias was intentionally retained to keep this PR small and minimize broad refactoring risk.
Moving forward, new/modified code should import explicitly:
- `AuthenticatedApiUtils` for protected endpoints
- `UnauthenticatedApiUtils` for public endpoints

## Maintainer question

@amritkv — would you prefer I do the explicit import migration (`ApiUtils` -> `AuthenticatedApiUtils`/`UnauthenticatedApiUtils`) in this PR now, or keep it as a separate follow-up refactor?

## Testing done (local)

- Verified session-expiry event flow and centralized redirect handling via `SessionStatusHandler`.
- Verified authenticated API path dispatches session-expiry event on `401`.
- Verified unauthenticated transport path remains available for public endpoints.

PR #1788, opened 2026-06-11
fix(frontend): Add use client files using ApiUtils

GMishx:
Add the 'use client' to app files which are using ApiUtils to make API calls. Also update check-client-directives.mjs with this check.

This PR includes files missed in https://github.com/eclipse-sw360/sw360-frontend/pull/1787

This PR also removes missed `useSession` and `getSession` from #1774


amritkv (approved):
LGTM !

PR #1798, opened 2026-06-16
Prefer runtime SW360_API_URL for server-side requests (kubernetes specific) 

CKnoppas:
NEXT_PUBLIC_SW360_API_URL is inlined at build time by Next.js/Turbopack. Server-side code (e.g. auth.service.ts) uses this URL to reach the backend, but the baked-in value is not reachable from inside the pod — in Kubernetes, .localhost TLDs resolve to 127.0.0.1 (RFC 6761), causing ECONNREFUSED on login. Unlike docker-compose, where port mapping makes localhost:8080 work from within containers, Kubernetes requires an internal service DNS name (e.g. http://sw360:8080) for pod-to-pod communication.

Prefer the non-prefixed SW360_API_URL env var at runtime so Kubernetes deployments can set the internal service URL for server-side calls. On the client side, SW360_API_URL is undefined (not exposed to browsers without the NEXT_PUBLIC_ prefix), so it falls back to the build-time NEXT_PUBLIC_SW360_API_URL — preserving existing browser behavior through the ingress.


amritkv (approved):
LGTM !

PR #1810, opened 2026-06-25
Toggle SBOM Import and Update Backend Config

deo002:
1. Added logic to enable/disable import/export sbom buttons on project details page based on backend config.
2. Right after updation of the backend config, the frontend still used the old config until the timely 15 min refresh. Added a refresh just after update.


amritkv (approved):
LGTM !

PR #1811, opened 2026-06-25
fix(docker): Remove nouveau overrides for h2c

GMishx:
Since Nouveau now has the connection type of http2 bydefault from version 3.5.2 onwards (https://github.com/apache/couchdb-docker/pull/287), the custom override is not required anymore.


amritkv (approved):
LGTM !

PR #1813, opened 2026-06-25
fix(licenseclearing): Remove redundant api calls in license clearing

deo002:
There are two sets of api calls being made for the same data, one for tree view and one for list view. The api calls were moved to a parent component and the data was passed to the child to avoid a set of redundant api calls.


amritkv (approved):
LGTM !

Issue #1815, opened 2026-06-26
Current Vendor not Selected in Vendor Modal in edit project and component page and create vendor bottom is disabled in add vendor screen


Farooq-Fateh-Aftab:
### Description

The current vendor does not come pre-selected because vendor id is not sent from backend. 

#### How to reproduce

Check edit page of components/projects.

amritkv:
Closing this issue as the PR https://github.com/eclipse-sw360/sw360-frontend/pull/1816 is merged.

PR #1816, opened 2026-06-26
fix(Project): Fixed vendor selection upon edit

Farooq-Fateh-Aftab:
### Summary
This PR fixes vendor selection consistency in project edit and vendor dialog flows.

- Preserves vendor identity reliably by preferring `vendorId` and falling back to vendor object id when needed.
- Improves vendor matching logic in the vendor search dialog to compare stable identifiers first, then use safe field-based fallback.
- Syncs dialog selected vendor state when incoming vendor prop changes.
- Ensures vendor id is always set in project payload using link id fallback to direct vendor id.
- Removes unnecessary disabled state on the create vendor submit button.

### Issue Link
Issue:  #1815 

How this solves it:
- Fixes cases where vendor selection was not retained correctly during project edit.
- Prevents mismatch when vendor data comes from different shapes (`_links.self.href` vs `id`).

### Suggest Reviewer
@GMishx @deo002 

### How To Test?
1. Open a project that already has a vendor assigned.
2. Go to Edit Project and verify the existing vendor appears selected.
3. Open the vendor search dialog and verify the correct row is pre-selected.
4. Select a different vendor, save, and verify the updated vendor persists after reload.
5. Create/select vendor via the add-summary flow and verify payload contains a valid `vendorId`.
6. Check that creating a vendor is still possible from the vendor dialog flow.

### Backend Related PR
https://github.com/eclipse-sw360/sw360/pull/4311

### DIFF
--- a/src/app/[locale]/admin/vendors/components/AddVendor.tsx
+++ b/src/app/[locale]/admin/vendors/components/AddVendor.tsx
@@ -68,7 +68,6 @@ export default function AddVendor(): JSX.Element {
                         <button
                             type='submit'
                             className='btn btn-primary col-auto me-2'
-                            disabled={status !== 'authenticated'}
                         >
                             {t('Create Vendor')}
                         </button>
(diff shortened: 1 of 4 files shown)


GMishx (approved):
LGTM

PR #1818, opened 2026-06-29
fix(docker): Fix the API token salt docs

GMishx:
Fix the generation of API Token salt in Docker setup. Most changes are only in docs.

Backend PR https://github.com/eclipse-sw360/sw360/pull/4315


deo002 (approved):
LGTM!

PR #1835, opened 2026-07-07
fix(search): Fix pagination on search page

deo002:
- Fix pagination on Search Page
- The error "The given page index is bigger than the actual number of pages" was being thrown when I navigate to a page further up and then change the page size. For example, let the total no. of page entries be 200. Let the page size be 10. So, when I go to page 4 and then change the page size to 100, I get the error above as page 4 does not have any entries to show. Fixed this as well.


amritkv (approved):
LGTM !

PR #1838, opened 2026-07-07
feat(language): Persist UI language preference

### LINKED ISSUE #1837: Persist UI Language
### Description

Persist the user's language settings for longer terms.

#### Steps followed and expected result

* When a user from non-English speaking country opens new UI, they get the interface in their language, which is good.
* After they login, they change their language. And it works as expected.
* But when they close the session and start a new session, they get the interface again in their native language. This is not a desired behavior.

#### Expected results

* User's selected language should be stored for long term so next time they start a new session, their language does not change back.
* Potentially use the `sw360user` to store the preferred language when the user changes it from the frontend. Next time they login, use the value to set the interface's language.

amritkv:
### Description

Users from non-English locales who changed their UI language found it reverted to their browser's native language after closing and reopening the app. The selected language was stored in the `NEXT_LOCALE` cookie, which — with this project's `localePrefix: 'never'` setup — defaulted to a session cookie (cleared on browser close). On the next visit, locale resolution fell back to the Accept-Language header, undoing the user's choice.

### Current Implantation (Temp)

Set an explicit `maxAge` on the locale cookie in `routing.ts`, converting `NEXT_LOCALE` from a session cookie into a persistent one (1 year validity).

This is the officially recommended approach in the `next-intl docs` for the `localePrefix: 'never'` mode. No backend, auth, or API changes required.

### Known limitations

The preference is stored client-side, so it does not follow the user across:

 - A different device
 - A different browser (Chrome → Firefox)
 - Cleared cookies / privacy tools

### Robust resolution (backend)

True cross-device persistence requires storing the preference on the user account:

1. Add a `preferredLanguage` field to the `sw360user` model (backend, Thrift + REST)
2. Return it on `GET` to `/users/profile` and accept it on `PATCH` to  `/users/profile`
3. Frontend: read it at login (seed the cookie from the account) and write it on language change

Closes : https://github.com/eclipse-sw360/sw360-frontend/issues/1837

### DIFF
--- a/src/i18n/routing.ts
+++ b/src/i18n/routing.ts
@@ -18,6 +18,11 @@ export const routing = defineRouting({
     localePrefix: {
         mode: 'never',
     },
+    // Persist the user's selected language across sessions in the same browser.
+    // Currently the maxAge is set to one year, which means the preference will be remembered for one year.
+    localeCookie: {
+        maxAge: 60 * 60 * 24 * 365,
+    },
 })
 
 export const { Link, redirect, usePathname, useRouter } = createNavigation(routing)


deo002 (approved):
LGTM!

PR #1839, opened 2026-07-07
fix(Project): Fixed info message on edit project based on user role

amritkv:
Fixed moderation request info message on edit project based on user role as detailed below :

 - The info banner will show as : `You are editing the original document` when
     - If the current use role is `ADMIN`
     - If the current user is the creator of the project
 - The info banner will show : `You will create a moderation request if you update` for all the other logged in users and user roles.


deo002 (approved):
LGTM!

PR #1841, opened 2026-07-08
fix(cd): Fix docker compose setup

GMishx:
1. Fix the environment variables.
2. Move more sensitive variables to secrets.
3. Update README_DOCKER.md
4. Make changes for new couchdb and couchdb-nouveau 3.5.2.1 with fixes for data directory and index upgrader.
5. New entrypoint for sw360-frontend to set the `NEXTAUTH_SECRET` and the `SW360_FRONTEND_KC_SECRET`
6. Upgrade variables for Keycloak to utilize v2 hostname features.
7. Simplify the docker compose setup by having only `docker-compose.yml` not two with `docker-compose-full.yml`


README_DOCKER.md
```diff
@@ -1,107 +1,311 @@
...
+| `https://localhost/resource` | SW360 REST API |
+| `https://localhost/authorization` | SW360 OAuth2 endpoints |
+| `https://localhost/kc` | Keycloak |
+
+The default certificate is a self-signed snakeoil cert — your browser will warn.
+Replace it for anything beyond local testing (see
+[Nginx (Reverse Proxy)](#nginx-reverse-proxy)).
+
+After this point, you should be good to use the stack with basic auth and
+visit the frontend at <https://localhost/>. The default credentials are `admin`
```
deo002: setup@sw360.org and sw360fossie

docker-compose.yml
deo002: what is 11311 port for?

Dockerfile
```diff
@@ -14,11 +14,11 @@ FROM node:${VARIANT} AS build
...
 
-# define run-time args
+# define build-time args
 ARG NEXT_PUBLIC_SW360_API_URL
 ARG NEXT_PUBLIC_SW360_AUTH_PROVIDER=sw360basic
 
-# define build-time args
+# define run-time args
 ARG NEXTAUTH_URL=http://localhost:3000
 ARG AUTH_SECRET=mysecret
```
deo002: AUTH_SECRET is not used anywhere, nor do I see it in the docs: https://next-auth.js.org/configuration/options. Could you please remove it?

config/front-end/.env.frontend
```diff
@@ -6,14 +6,12 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 # License-Filename: LICENSE
-# Change this to strong secret
-NEXTAUTH_SECRET=secret
 # Where the frontend will run
-NEXTAUTH_URL=http://localhost:3000
+NEXTAUTH_URL=https://localhost
```
deo002, line 10: We also need another runtime env var SW360_API_URL=http://sw360:8080 for this to work.

In src/utils/env.ts, we have the variable SW360_API_URL. In browser bundles this takes the value of NEXT_PUBLIC_SW360_API_URL as SW360_API_URL is not present there. In frontend's server, this should take the value http://sw360:8080 (SW360_API_URL should be set to this val) for api calls to auth. But now, it was taking the value of NEXT_PUBLIC_SW360_API_URL, leading to errors.

README_DOCKER.md
```diff
@@ -1,107 +1,311 @@
...
+```
+
+Then build the frontend image:
+
+```sh
+docker compose -f docker-compose.yml build sw360-frontend
+```
+
+### 3. Start the Stack
+
```
deo002, line 143: I think it will be a good idea to add a 4th point to keep the .env.frontend in sync with these values as well, just in case the developer forgets to edit them.

deo002 (changes requested):

README_DOCKER.md
GMishx: Fixed.

docker-compose.yml
GMishx: That's an artifact from before. Removed it.

Dockerfile
GMishx: Yeah because it is `NEXTAUTH_SECRET`. Fixed.

config/front-end/.env.frontend
GMishx: Okay, this runtime variable was added in #1798 recently. I will create a separate PR for this. The current documentation and everything relies on the old `NEXT_PUBLIC_SW360_API_URL` buildtime variable. Which should still work IMO

https://github.com/eclipse-sw360/sw360-frontend/blob/2da85a226bcab64e3c4c8bbfdd56e1f8f80f4c60/src/utils/env.ts#L13

README_DOCKER.md
GMishx: In that case, you'll forget to read the docs too :-)

deo002 (approved):
LGTM!

Issue #1842, opened 2026-07-08
Improve auto complete input


amritkv:
Can't focus popups of "suggestion lists" with keyboard

To reproduce:

* Go to `Add Component` page
* Select "Categories" field and start to type "li". Now we have to use the mouse to select "library", neither cursor down (nothing happens) nor tab key works

This seems to only happen for normal input fields with a popup of suggested content, so for example the "Component Type" field works as expected (using arrow up/down, we can select the item from the list), but for "Categories" or the "external id key" field, we're not able to select suggested values with the keyboard.

Note : Also propagate this improvement for all the other pages, wherever applicable

saad-mohammed1:
Hi, I'm working on this issue and will submit a pull request once the implementation and testing are complete.

amritkv:
Closing this issue as the PR is merged here : https://github.com/eclipse-sw360/sw360-frontend/pull/1843

PR #1843, opened 2026-07-09
fix(ui): Add keyboard navigation support for suggestion lists (issue: #1842)

saad-mohammed1:
## Summary

This PR adds keyboard navigation support for suggestion lists across the UI to improve accessibility and usability.

## Changes

- Added keyboard navigation using Arrow Up and Arrow Down keys
- Added Enter and Tab support for selecting suggestions
- Added Escape support to close the suggestion list
- Applied the improvement to the shared `SuggestionBox` component so it is available wherever it is used

### DIFF
--- a/src/app/[locale]/components/add/components/AddComponent.tsx
+++ b/src/app/[locale]/components/add/components/AddComponent.tsx
@@ -74,11 +74,9 @@ function AddComponent(): ReactNode {
         blog: '',
     })
 
-    // Configs from backend
-    const componentExternalIdSuggestions =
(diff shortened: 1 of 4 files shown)


amritkv:
Hey @saad-mohammed1 !
Tested the functionality and found one small bug.
At the component add/edit page, the navigation keys are working. The point here is to note, scroll is working in background but no change of scroll in UI.

Attached a screenshot :

image

saad-mohammed1:
Hi @amritkv Thanks for testing!
 I've fixed the scrolling issue now. The highlighted suggestion is kept **in view** while navigating with the keyboard. I've updated the PR with the fix.

amritkv (approved):
LGTM !

PR #1854, opened 2026-07-13
fix(UI): Show ECC pill in red when open ECC exists

nikkuma7:
1: Login sw360
2: Navigate to Projects
3: Search for ex: XYZ release
4: ECC field is red if ECC state of one of the components is open
5: Add the total number of components with open ECC state in the expected result


amritkv (approved):
LGTM !

PR #1874, opened 2026-07-20
fix(Vendor): Vendor preselection in vendor search in the project component edit page

Farooq-Fateh-Aftab:
This PR fixes the vendor search popup (`VendorDialog`) so that the currently saved vendor is shown and pre-selected when the dialog is opened from an edit screen.

Previously:

- When a component/project/release already had a saved vendor, clicking the vendor text box in the edit screen opened the **Search Vendor** popup with an empty results table. The saved vendor was neither listed nor highlighted, so the user could not see which vendor was currently set.

Changes:

- `src/components/sw360/SearchVendorsModal/VendorDialog.tsx`
  - Added an optional `overrideSearchText` parameter to `searchVendor` so a search can run immediately with a given term without waiting for the `searchText` state to update.
  - Added a `useEffect` on `show` that, when the dialog opens with a saved vendor, pre-selects that vendor and runs a search by its full name so the saved vendor appears in the results table and its radio button is pre-selected.
  - The visible search text box is intentionally left empty (not prefilled), while the saved vendor still loads in the list.

Because `VendorDialog` is shared, this fix applies to the component (`GeneralInfoComponent`), project (`ProjectAddSummary/GeneralInformation`), and release (`ReleaseSummary`) edit screens.

Issue: 

### Suggest Reviewer

> @amritkv , @deo002 

### How To Test?

1. Run the app and log in.
2. Open a component (or project/release) that already has a saved default vendor and go to its edit screen.
3. Click the vendor text box to open the **Search Vendor** popup.
4. Verify the saved vendor is listed in the results table and its radio button is already selected.
5. Verify the search text box at the top of the popup is empty (not prefilled).
6. Verify searching, resetting, selecting a different vendor, and clearing the vendor still work as before.

### DIFF
--- a/src/app/[locale]/admin/bulkreleaseedit/components/ReleaseEditPage.tsx
+++ b/src/app/[locale]/admin/bulkreleaseedit/components/ReleaseEditPage.tsx
@@ -44,7 +44,7 @@ function UpdateReleaseModal({
     useEffect(() => {
         setVendor(release?.vendor ?? {})
     }, [
-        release,
+        release?.id,
     ])
 
     function handleClose() {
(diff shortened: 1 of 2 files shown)


deo002 (approved):
LGTM! Tested, working as expected.

PR #1879, opened 2026-07-21
fix(Attchment): Fixed attachment update status keyword

amritkv:
Fixed the enum for attachment status `Not Checked` while updating an Attachment.

Currently the enum `NOT_CHECKED` is sent for the status value `Not Checked` in the payload of update attachments but backend expects it as `NOTCHECKED`.

This PR fixes this bug.

Linked backend PR : https://github.com/eclipse-sw360/sw360/pull/4374


GMishx (approved):
LGTM

PR #1880, opened 2026-07-22
Fix issues in admin vendor page

deo002:
- Add sorting be short name to vendors table
- On pressing enter after adding data in quick filter, the search text clears and table is set to the default one. Fixed this.


GMishx (approved):
LGTM

PR #1882, opened 2026-07-22
fix(footer): Fixed rest api doc url at footer

amritkv:
Temp fix for the `Rest API Doc` url with a new variable.


amritkv:
Looking for better approach than temp solution.

PR #1884, opened 2026-07-23
Revert "fix(UI): Show ECC pill in red when open ECC exists"

GMishx:
Reverts eclipse-sw360/sw360-frontend#1854 since the backend branch is not yet merged and this causes confusion for the user.


amritkv (approved):
Agreed to it.

PR #1886, opened 2026-07-23
Fix issues caused in urls of user detail and edit page by special symbols in email

deo002:
Some users have emails as ids as urls such as 'http://localhost:3000/admin/users/details/user@sw360.org'. Special symbols in email such as '@' and '.' cause issues in email parsing of nextjs. While '@' can be replaced with %40, we have no replacement of '.', so 'http://localhost:3000/admin/users/details/user%40sw360.org' does not work either. The only option is to make it a query param rather than query slug. The license detail and edit pages also use query params for ids due to the same reason(presence of '.'s and other symbols in license names).

So, now, the url is 'http://localhost:3000/admin/users/edit?id=user%40sw360.org'

image


GMishx (approved):
LGTM

PR #1889, opened 2026-07-23
fix(project): Permission check while editing project

amritkv:
Fixes in this PR :

 - Added permission check while routing from `Project Detail` page to `Edit Project` page
 - Fixed API response status code-based banner message while updating a project


GMishx (approved):
LGTM

PR #1890, opened 2026-07-23
fix(project): Show error message when license report download fails

rudra-superrr:
Description: Show error message when license report download fails.


amritkv (approved):
LGTM !

PR #1891, opened 2026-07-23
fix(project): Handled forbidden case for save attachments

amritkv:
This PR fixes below issues at save attachments:
 - If user doesn't have save attachment usage permission
 - Handled errors from backend while report generation


GMishx (approved):
Good to go.

PR #1892, opened 2026-07-23
fix(releases): Get releases from api call instead of embedded object due to number mismatch

deo002:


GMishx (approved):
Good work around. But need a better implementation and fix in backend for the embedded releases.

PR #1893, opened 2026-07-24
fix(dept): Allow depts other than the ones in list

deo002:
### Description
Currently, the new UI does not allow writing other department/group name in the Department Selector. This is undesired behavior. This field should be modifiable for the user.

I have made it modifiable.

### Test
image

image


amritkv (approved):
LGTM !

PR #1894, opened 2026-07-24
Make all releases and moderators visible

deo002:
In Obligations tables in Project details/edit page, not all releases were visible after opening the dropdown. Same with moderation requests page. Fixed this issue.

image

image

image


GMishx (approved):
Good to go.

PR #1896, opened 2026-07-24
Handle no content response in link releases modal correctly

deo002:
Handle 204 response correctly in Link Releases Modal

image


GMishx (approved):
Good to go.

PR #1897, opened 2026-07-24
fix(Component): Refesh subscribe button name and color

Farooq-Fateh-Aftab:
### Summary

On the **Component Detail** page, clicking the **Subscribe** button correctly performed the subscription on the backend, but the button did not update its label (`Subscribe` → `Unsubscribe`) or its color (`outline-success` → `outline-danger`) until the page was reloaded.

### Suggested Reviewer
@GMishx @deo002 @amritkv

### DIFF
--- a/src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx
+++ b/src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx
@@ -62,7 +62,6 @@ const DetailOverview = ({ componentId }: Props): ReactNode => {
     const [subscribers, setSubscribers] = useState<Array<string>>([])
     const [changeLogId, setChangeLogId] = useState('')
     const [changelogTab, setChangelogTab] = useState('list-change')
-    const [refreshSubscriptions, setRefreshSubscriptions] = useState(false)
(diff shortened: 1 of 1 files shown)


src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx
```diff
@@ -164,7 +163,16 @@ const DetailOverview = ({ componentId }: Props): ReactNode => {
     const handleSubcriptions = async () => {
         try {
             await ApiUtils.POST(`components/${componentId}/subscriptions`, {})
```
Comment (amritkv) [inline src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx]: Hey @Farooq-Fateh-Aftab !
Here we're not checking the status/response for the POST call to subscription API.
Irrespective of the POST call status, for example if it fails, the next code block still goes to re-fetch the component details.

amritkv (changes requested):

src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx
Comment (Farooq-Fateh-Aftab) [reply src/app/[locale]/components/detail/[id]/components/DetailOverview.tsx]: Hi @amritkv 
Comment addresed

amritkv (approved):
LGTM !

PR #1915, opened 2026-07-30
fix(ci): Temp fix to use TS `v7.x` with TS `v6.x`

amritkv:
# Introducing TypeScript 7.x native compiler with TypeScript 6.x

## Problem

Dependabot PR #1907 bumped `typescript` from `6.0.3` to `7.0.2`. TypeScript 7 is a complete **rewrite in Go** and ships a **different package layout** — it no longer includes `lib/typescript.js`, which Next.js 16.x relies on for its build-time type check.

In more detailed terms: Next.js's build process relies on TypeScript's programmatic API (the old JavaScript typescript package) to run type-checking under the hood. Because TypeScript 7.0 completely ported the compiler to a native Go binary and shipped without a stable programmatic API, Next.js's internal `require('typescript')` call throws an exception.

As a result, `next build` failed in CI with:

```
It looks like you're trying to use TypeScript but do not have the required package(s) installed.
Next.js build worker exited with code: 1
```

In short: **TypeScript 7 is not yet a drop-in replacement**, because tools like Next.js, `ts-node`, `tsc-files`, and `prettier-plugin-organize-imports` still depend on the stable `6.x` JavaScript API.

TypeScript 7.0 release announcement blog by Microsoft : https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

## Solution

Run **TypeScript 6.x and the TypeScript 7 native compiler side by side**, each with a clear role:

| Package | Role | Used by |
|---------|------|---------|
| `typescript@^6.0.3` (`tsc`) | Stable **API provider** | VS Code, `ts-node`, `tsc-files`, prettier plugin |
| `@typescript/native-preview` (`tsgo`) | Fast **type-checker** | `pnpm type-check` (local dev + CI) |

Type-checking is now a fast, dedicated CI/dev gate powered by the native `tsgo` compiler, while `next build` focuses on bundling (Turbopack).

## Changes

- **`package.json`**
  - Added dev dependency (pinned): `@typescript/native-preview@7.0.0-dev.20260707.2`
  - Added script: `"type-check": "tsgo --noEmit"`
  - Kept `typescript` on `^6.0.3` as the stable API provider
- **`next.config.ts`**
  - Set `typescript.ignoreBuildErrors: true` — type errors are now caught earlier by the faster `type-check` gate, so the slow build-time TypeScript pass is skipped. This also prevents a future TS 7 bump from breaking `next build`
- **`.github/workflows/build_deploy.yml`**
  - Added a **Type check** step (`pnpm type-check`) before the build step — a real type gate that did not exist before
- **`.github/dependabot.yml`**
  - Added an ignore rule for `typescript` `7.x` so Dependabot stops re-proposing the incompatible bump until Next.js supports the native compiler
- **`global.d.ts`**
  - Added a single triple-slash reference so static asset imports (`*.svg`, `*.png`, ...) type-check independently of the build:
    ```ts
    /// <reference types="next/image-types/global" />
    ```
  - Critical for CI: without it the `type-check` step fails
  - Root cause: SVG/image module types come from Next's auto-generated **`next-env.d.ts`**, and only created by `next dev` / `next build`. Because the `type-check` step now runs **before** the build, that file does not exist yet in CI, so the `declare module '*.svg'` types are missing.

## Benchmarking

- Benchmark on this codebase (same `tsconfig.json`):
  - `tsgo` (7.x native): **~8.2s**
  - `tsc` (6.x JS): **~37.9s**
  - ≈ **4.6× faster** type-checking

## Benefits

- ⚡ Significantly faster type-checking via the native Go compiler
- ✅ Adds a dedicated type gate on CI (previously only `next build` caught type errors)
- 🔒 Zero risk — `6.x` remains the source of truth; `tsgo` is additive
- 🧩 No source-code changes — same `tsconfig.json`, config-only changes
- 🚀 Clear migration path — once TypeScript 7 stabilizes and Next.js supports the native compiler, drop `@typescript/native-preview`, bump `typescript` to `7.x`, and remove the Dependabot ignore rule

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,14 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      # Temp fix, until typescipt 7.x is supported by Next.js.
+      # TypeScript 7.x is the native (Go) rewrite and ships a different package
+      # layout (no lib/typescript.js). Next.js 16.x build-time type checking
+      # cannot load it and fails with "do not have the required package(s)
+      # installed". Stay on the JS-based 6.x line until Next.js supports it.
+      - dependency-name: typescript
+        versions: ['7.x']
 
   - package-ecosystem: docker-compose
     directory: /
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -78,6 +78,11 @@ jobs:
             echo '```'
           } >> $GITHUB_STEP_SUMMARY
 
+      - name: Run Type checks
+        shell: bash
+        run: |
+          pnpm type-check
+
       - name: Run build
         shell: bash
         run: |
--- a/global.d.ts
+++ b/global.d.ts
@@ -7,6 +7,14 @@
 // SPDX-License-Identifier: EPL-2.0
 // License-Filename: LICENSE
 
+// Ambient declarations for static asset imports (*.svg, *.png, *.jpg, ...).
+// Next.js normally provides these via the auto-generated, git-ignored
+// `next-env.d.ts`, which only exists after `next dev` / `next build`. The
+// standalone `type-check` step (tsgo) runs before the build in CI, so we
+// reference Next's image types directly here to keep type-checking independent
+// of the build step (reusing the exact same declarations avoids type drift).
+/// <reference types="next/image-types/global" />
+
 type Messages = typeof import('./messages/en.json')
 declare interface IntlMessages extends Messages {}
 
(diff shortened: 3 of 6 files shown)


amritkv:
Closing this PR as this is creating complexity to use `TypeScript 7.x` along with `TypeScript 6.x`, just for speed only. Better let's wait for `NextJs` to be `TypeScript 7.x` compatible as `TypeScript 7.x` has not yet published its programmable APIs.

PR #1933, opened 2026-08-05
fix(Project): Fixed linked release loading on duplicate

amritkv:
This PR fixes the issues listed below:
 - Removed the `modifiedBy` and `modifiedOn` field from payload of duplicate project creation as it'll handled by backend
 - Made the `Create Project` button disabled until all the linked releases are loaded at duplicate project page
 - Introduced the `Processing...` info on the linked release table while the linked releases are loading


GMishx (approved):
LGTM

PR #1934, opened 2026-08-06
fix(UI): Resolve global search API call issue and display filtered result

nikkuma7:
This PR fixes the following:

1. Fixed global search submit flow to pass search text in URL query params.
2. Synced search page input with URL searchText so results load correctly from Navbar search.
3. Resolved issue where global search showed unfiltered results.

Issue Link

Issue: : #1935 

Suggest Reviewer
@GMishx @deo002 

How To Test?

1: Login sw360
2: open any page and put the value in global search textbox.
3: will get the filtered result in search tab.


amritkv (approved):
LGTM !

PR #1952, opened 2026-08-12
fix(moderation): Fix search and add sort in moderation requests

deo002:
In Moderation Requests page,
1. Added Server Side Pagination
2. Added Server Side Sorting
3. Updated Advanced Search to match changes in [feat/search/rework](https://github.com/eclipse-sw360/sw360/tree/feat/search/rework) branch


amritkv (approved):
LGTM !

PR #1956, opened 2026-08-12
fix(Project): Update multiple obligation

Farooq-Fateh-Aftab:
[//]: # (This program and the accompanying materials are made)
[//]: # (available under the terms of the Eclipse Public License 2.0)
[//]: # (which is available at https://www.eclipse.org/legal/epl-2.0/)
[//]: # (SPDX-License-Identifier: EPL-2.0)

### Summary

Previously, when saving a project, each obligation type (license, component, project, organisation) was updated via separate PATCH requests regardless of how many types were present. This caused only the last request to persist when multiple obligation types were modified together.

This fix buckets obligations by their `obligationType` before dispatching updates:
- If obligations of **multiple types** are present, all are sent in a single `PATCH projects/{id}/updateObligation?obligationLevel=all` request.
- If obligations of only **one type** are present, the existing type-specific endpoint is used (`updateLicenseObligation`, `updateObligation?obligationLevel=component/project/organization`).

This ensures that saving a project with mixed obligation types correctly persists all changes.

Issue: https://github.com/eclipse-sw360/sw360-frontend/issues/1955

### How To Test?

1. Open an existing project that has obligations of **more than one type** (e.g. both License and Component obligations).
2. Change the status/comment on obligations from different types.
3. Save the project.
4. Re-open the project and verify that all obligation changes across all types are persisted correctly.

Previously only one obligation type would be saved; after this fix all types should reflect the changes.

### Checklist

Must:
- [ ] All related issues are referenced in commit messages and in PR

### DIFF
--- a/src/app/[locale]/projects/edit/[id]/components/EditProject.tsx
+++ b/src/app/[locale]/projects/edit/[id]/components/EditProject.tsx
@@ -570,42 +570,86 @@ function EditProject({
                 ),
             ]
             if (Object.keys(obligations).length !== 0) {
+                const licenseObligations: Record<string, object> = {}
+                const componentObligations: Record<string, object> = {}
+                const projectObligations: Record<string, object> = {}
+                const organisationObligations: Record<string, object> = {}
+
                 for (const key in obligations) {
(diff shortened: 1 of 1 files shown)


amritkv (approved):
LGTM !

PR #1972, opened 2026-08-17
fix(release): Adding the clearing report column in release overview

rudra-superrr:
Adding the Clearing Report column in release overview table in Components page.

image


akshitjoshii:
Test successful. same behaviour as old backend.

image

amritkv (approved):
LGTM !

PR #1958, opened 2026-08-13
feat(Config): add REST token write-access preferences toggle

bibhuti230185:
## Summary

This PR introduces a new frontend configuration to control whether the Write Access checkbox is shown in Preferences > REST API Tokens.

### What was added
- New frontend config key:
  - `ui.rest.api.write.access.token.in.preferences.enabled`
- New toggle in Admin > Configurations > Frontend Configurations:
  - Enable Write Access option for REST API Tokens in Preferences
- Preferences token form behavior update:
  - Write Access checkbox is shown only when:
    - REST API Token generator is enabled (`ui.rest.apitoken.generator.enable`)
    - New preferences toggle is enabled (`ui.rest.api.write.access.token.in.preferences.enabled`)
  - If the checkbox is hidden, `WRITE` is removed from authorities before token generation.
- i18n updates:
  - Added localized label/help text entries for all supported locales.

## Files Changed
- `src/object-types/enums/UIConfigKeys.ts`
- `src/app/[locale]/admin/configurations/components/FrontEndConfigs.tsx`
- `src/app/[locale]/preferences/components/UserAccessToken.tsx`
- `messages/en.json` (and all 9 other locale files)
- `tests/e2e/preferences/preferencesPage.test.ts` *(deterministic ON/OFF E2E tests added; skipped until backend is deployed)*

## Suggested Reviewer
@amritkv @GMishx 

## Depends On Backend PR
https://github.com/eclipse-sw360/sw360/pull/4486
## Manual Test Steps

1. Open Admin > Configurations > Frontend Configurations.
2. Set Enable REST API Token generator = On.
3. Set Enable Write Access option for REST API Tokens in Preferences = On.
4. Open Preferences > REST API Tokens and confirm Write Access checkbox is visible.
5. Set Enable Write Access option for REST API Tokens in Preferences = Off.
6. Refresh Preferences and confirm Write Access checkbox is hidden.

PR #1976, opened 2026-08-18
fix(UI): Prevent empty table message from flashing before data loads.

nikkuma7:
## Summary
- Fixed incorrect table state order on initial page load across the app.
- Fix applied in the shared table component, so all tables and tabs are covered.
- Prevented superseded (aborted) API requests from resetting the loading state of the newer request.
## Root Cause
- Pages start their fetch after mount and flip showProcessing via setTimeout(..., 0), so on the first frames the table is empty while     showProcessing is still false. SW360Table could not distinguish "not loaded yet" from "loaded and empty", and rendered "No data available in table".
- On the Projects page a second effect resets pageableQueryParam on mount, which aborts the first request. The aborted request's finally block then ran setShowProcessing(false), clearing the loading state that the newer, still in-flight request owned.
## What Changed
Frontend (SW360Table): the processing overlay is kept until the table has stayed empty and idle for a short grace period, so the empty-state message can only render once a request has actually finished with no rows.
Frontend (ApiUtils): a request cancelled through the caller's own AbortSignal never settles, so its stale catch/finally handlers cannot reset loading flags or state already owned by the newer request.
## Issue:

Suggest Reviewer
@GMishx @amritkv .

How To Test?
How should these changes be tested by the reviewer?
Have you implemented any additional tests?

1. Open the Projects page (/projects) and reload it.
2. Verify the sequence is Processing → data, and that "No data available in table" never appears in between.
3. Repeat on /components, /licenses, /vulnerabilities and on tabbed tables (e.g. project details tabs).
4. Verify the empty state still works: search for something with no results and confirm "No data available in table" is shown after     the request completes.
5. Test on a throttled network and confirm the processing overlay stays visible until the response completes.
6. Verify pagination and page-size changes on a table that already has rows still keep the existing rows visible instead of flashing an empty state.

### DIFF
--- a/src/components/sw360/Table/Components.tsx
+++ b/src/components/sw360/Table/Components.tsx
@@ -7,13 +7,35 @@
 // SPDX-License-Identifier: EPL-2.0
 // License-Filename: LICENSE
 
+'use client'
+
 import { ColumnFiltersState, flexRender, Row, Table } from '@tanstack/react-table'
 import { useTranslations } from 'next-intl'
-import { ChangeEvent, Dispatch, Fragment, ReactNode, SetStateAction } from 'react'
+import {
+    ChangeEvent,
+    Dispatch,
+    Fragment,
+    ReactNode,
+    SetStateAction,
+    useDeferredValue,
+    useEffect,
+    useRef,
+    useState,
+    useSyncExternalStore,
+} from 'react'
 import { Dropdown, DropdownButton } from 'react-bootstrap'
 import { BiSort } from 'react-icons/bi'
 import { BsCaretDownFill, BsCaretRightFill, BsSortDown, BsSortDownAlt } from 'react-icons/bs'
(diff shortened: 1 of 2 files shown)


amritkv:
Hey @nikkuma7 !
The e2e test cases are failing, can you please have a look ?

nikkuma7:
> Hey @nikkuma7 ! The e2e test cases are failing, can you please have a look ?

If the backend's UI_REST_APITOKEN_WRITE_GENERATOR_ENABLE config resolves to false. the checkbox never renders — nothing to do with our table/api.util.ts or components.ts change.


Confirmed — this failure is completely unrelated to our table/API fix.

GMishx:
@nikkuma7 since #1958 is merged, please rebase your branch and the test cases should pass.

nikkuma7:
> @nikkuma7 since #1958 is merged, please rebase your branch and the test cases should pass.

rebase done.

amritkv:
Closing this PR as the same fix already merged in the PR : #2053

PR #1983, opened 2026-08-21
fix(Release): Ecc info hover text

rudra-superrr:
Issue: The _(i)_ info popup button text in the "ECC Details" tab in Edit Release page has its pop up text written next to the button, not in the pop-up.

before:
image

after:
image


GMishx (approved):
LGTM

PR #1984, opened 2026-08-21
fix(Fossology): Updating fossology config error msg

rudra-superrr:
Description: When saving the FOSSology configuration with an invalid URL, the backend returns a 400 response with a descriptive error message (e.g. "Base URL for API must be 'v2'."), but the frontend was discarding the response body and always showing a generic "Fossology configuration update failed" toast. The fix reads the error detail from the response JSON and displays it directly in the error toast, falling back to the generic message if no detail is available.

Before:
image

After:
image
image
image


GMishx (approved):
LGTM

PR #1997, opened 2026-08-25
fix(Project): Project Mainline State column

rudra-superrr:
Description: 
Project Mainline State column is always empty in license clearing table in project page.

Before:
image

After:
image


amritkv (approved):
LGTM !

PR #2000, opened 2026-08-26
chore(ci): Fixed hardcoded pnpm version

amritkv:
The `pnpm v11` was automatically resolving to `pnpm v11.12.0` which had bugs. So, hardcoded the `pnpm` version to `v11.11.0`.
But as of now, the bugs are fixed in upstream, so reverted back to major `pnpm v11`.


GMishx (approved):
LGTM

PR #2004, opened 2026-08-27
chore(Footer): Added deployed frontend repo info

amritkv:
Added deployed frontend repo info at the footer. Attached a screenshot herewith:

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 # syntax=docker/dockerfile:1.4
 
 # Copyright (c) Helio Chissini de Castro, 2023. Part of the SW360 Frontend Project.
+# Copyright (C) Siemens AG, 2026. Part of the SW360 Frontend Project.
 
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
@@ -17,6 +18,13 @@ ARG NPM_CONFIG_REGISTRY=https://registry.npmjs.org/
 # define build-time args
 ARG NEXT_PUBLIC_SW360_API_URL
 ARG NEXT_PUBLIC_SW360_AUTH_PROVIDER=sw360basic
+
+# Frontend build metadata — .git is excluded via .dockerignore, so pass these
+# from CI (e.g. `docker build --build-arg SOURCE_BRANCH=$(git rev-parse --abbrev-ref HEAD) ...`).
+ARG SOURCE_BRANCH
+ARG SOURCE_COMMIT
+ENV SOURCE_BRANCH=${SOURCE_BRANCH}
+ENV SOURCE_COMMIT=${SOURCE_COMMIT}
 
 # define run-time args
 ARG NEXTAUTH_URL=http://localhost:3000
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -18,6 +19,13 @@ ARG NPM_CONFIG_REGISTRY=https://registry.npmjs.org/
 ARG NEXT_PUBLIC_SW360_API_URL
 ARG NEXT_PUBLIC_SW360_AUTH_PROVIDER=sw360basic
 
+# Frontend build metadata — .git is excluded via .dockerignore, so pass these
+# from CI (e.g. `docker build --build-arg SOURCE_BRANCH=$(git rev-parse --abbrev-ref HEAD) ...`).
+ARG SOURCE_BRANCH
+ARG SOURCE_COMMIT
+ENV SOURCE_BRANCH=${SOURCE_BRANCH}
+ENV SOURCE_COMMIT=${SOURCE_COMMIT}
```
GMishx, line 27: Not a good idea to read these generated info as args. It should be calculated curing the build itself.

You can check how it is done in backend:
1. We make sure that `git` is installed in the build phase
    https://github.com/eclipse-sw360/sw360/blob/a031c1b9940ba8256be6a9222939daff3f5d3a0d/Dockerfile#L31
2. Mount the entire directory for build, which brings the `.git` foler
   https://github.com/eclipse-sw360/sw360/blob/a031c1b9940ba8256be6a9222939daff3f5d3a0d/Dockerfile#L48
3. Use a plugin for maven to get this info in a property file
   https://github.com/eclipse-sw360/sw360/blob/a031c1b9940ba8256be6a9222939daff3f5d3a0d/rest/resource-server/pom.xml#L224-L255
4. The property file is then loaded when required
    https://github.com/eclipse-sw360/sw360/blob/a031c1b9940ba8256be6a9222939daff3f5d3a0d/rest/resource-server/src/main/java/org/eclipse/sw360/rest/resourceserver/Sw360ResourceServer.java#L129
5. This eliminated loading anything as an ARG during image build
   https://github.com/eclipse-sw360/sw360/blob/a031c1b9940ba8256be6a9222939daff3f5d3a0d/.github/workflows/sw360_container.yml#L150-L163

Adding these parameters as builtime arg will make the image build more complex and someone might forget and wonder what went wrong.

next.config.ts
```diff
@@ -7,13 +8,70 @@
...
+ * Values are inlined into the client bundle via the Next.js `env` field.
+ */
+function resolveFrontendBuildInfo(): Record<string, string> {
+    const pkg = JSON.parse(readFileSync(resolve(process.cwd(), 'package.json'), 'utf-8')) as {
+        version: string
+    }
+
+    const tryGit = (args: string): string | undefined => {
+        try {
+            const out = execSync(`git ${args}`, {
```
GMishx, line 39: Please never call `git` at run time. Only buildtime is safe.

GMishx (changes requested):

PR #2029, opened 2026-09-03
fix(Fossology): Fossology clearing process stuck at step 2

rudra-superrr:
Problem
When FOSSology clearing process completes with report download disabled (2 steps instead of 3), the frontend UI:

- Shows incorrect progress percentage (66.64% instead of 100%)
- Displays wrong completion message ("Report generation done" instead of "Scanning source done")
- Remains stuck at "in progress" status even after backend returns SUCCESS

Root Cause
Frontend logic hardcoded assumptions about:

- Always expecting 3 process steps
- Always showing the same completion message regardless of actual step count
- Not handling the dynamic 2-step completion scenario

Related backend PR: https://github.com/eclipse-sw360/sw360/pull/4565


GMishx (approved):
Good to go.

PR #2033, opened 2026-09-04
fix(Attachment): Fix drag and drop attachment

rudra-superrr:
Description: now user can drag and drop multiple files while uploading attachments.


GMishx (approved):
LGTM

PR #2035, opened 2026-09-04
fix(project): Linked to projects bug

rudra-superrr:
Issue summary:

- The Link to Projects flow was linking in the wrong direction.
- Expected: from Project A, selecting Project B should make A a child of B.
- Actual: B was becoming child of A.

What was changed:

1. Direction fix in submit logic:

- In LinkProjectsModal.tsx:389, linking now updates each selected target project and injects the current project into that target’s linkedProjects map.
- This preserves existing target links while adding the new child relation.

2. Payload conversion helper:

- LinkProjectsModal.tsx:354 converts backend linkedProjects entries into the PATCH object shape, so existing links are not overwritten.

3. Update-mode selection cleanup:

- LinkProjectsModal.tsx:62 clears preselected entries in UPDATE mode to avoid stale/extra names in the success message.

4. Success message formatting:

- LinkProjectsModal.tsx:369 formats display as Name(version) when version exists.
- LinkProjectsModal.tsx:437 now lists all selected target projects explicitly (no +N more), matching the requested sentence style.

### DIFF
--- a/src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
+++ b/src/app/[locale]/projects/detail/[id]/components/ProjectDetailTab.tsx
@@ -175,6 +175,8 @@ export default function ViewProjects({ projectId }: { projectId: string }): JSX.
                 })
                 setProjectPayload({
                     id: projectId,
+                    name: project.name ?? '',
+                    version: project.version ?? '',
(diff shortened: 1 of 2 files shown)


amritkv (dismissed):

GMishx (approved):
Good to go now.

PR #2049, opened 2026-09-09
fix(Project): Added missing comment at license clearing tab

amritkv:
Added missing comment at `License Clearing` tab under Project detail page.


src/utils/common.utils.ts
```diff
@@ -330,6 +330,13 @@ const extractEmailsAndFullNamesFromUsers = (
     )
 }
 
+const truncateShortText = (item: string | undefined): string => {
```
deo002, line 333: There's a function for this: https://github.com/siemens/sw360-frontend/blob/72adb6d16bd29a0de4591c7dfb97313253ee3279/src/utils/common.utils.ts#L249

src/app/[locale]/projects/detail/[id]/components/TreeView.tsx
Comment (deo002) [inline src/app/[locale]/projects/detail/[id]/components/TreeView.tsx:432]: we can add ```comment: l.comment``` here as well

Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/TreeView.tsx:432]: With this, we can shorten the column rendering logic to just showing row.original.node.entity.comment ?? ''

Comment (deo002) [reply src/app/[locale]/projects/detail/[id]/components/TreeView.tsx:432]: Similar thing is already done in ListView

deo002 (changes requested):

deo002:
The font also seems a little different, or is it just me?
image

amritkv:
> The font also seems a little different, or is it just me? image

Font is same, just increase the font size. Now it looks similar to other texts in the page.

image

src/utils/common.utils.ts
amritkv: Reused the existing one.

src/app/[locale]/projects/detail/[id]/components/TreeView.tsx
Comment (amritkv) [reply src/app/[locale]/projects/detail/[id]/components/TreeView.tsx:432]: Yupp, added the comment here as well.

src/app/[locale]/projects/detail/[id]/components/TreeView.tsx
```diff
@@ -821,18 +823,36 @@ export default function TreeView({
...
+                    if (row.original.node.type !== 'release') return null
+
+                    const parentRow = row.getParentRow()
+                    const parentNode = parentRow?.original.node
+
+                    let comment = ''
+                    if (!parentNode) {
+                        // Top-level release (direct release of the current project) — no parent row
+                        comment = row.original.node.entity.comment ?? ''
+                    } else if (parentNode.type === 'project') {
```
Comment (deo002) [inline src/app/[locale]/projects/detail/[id]/components/TreeView.tsx]: We don't need this right? As we have added l.comment to the release row object itself

deo002 (changes requested):

src/app/[locale]/projects/detail/[id]/components/TreeView.tsx
Comment (amritkv) [reply src/app/[locale]/projects/detail/[id]/components/TreeView.tsx]: Good catch.  It is not required.

deo002 (approved):
LGTM!

PR #2053, opened 2026-09-10
Fix the brief empty table flash before data shows after processing

deo002:
Currently, when we visit a page with a table, it initially shows 'Processing...', after which, for a brief interval of time, empty table is shown, only after which the table with data shows up.

The empty table is shown due to a race condition on the showProcessing variable in many pages. For example, in src/app/[locale]/components/components/ComponentsTable.tsx, there are the following useEffects:
1. One which loads data from the api.
2. Another one which changes url params.

What happens when useEffect (1) and (2 )run together is, (1) sets showProcessing to true, simultaneously (2) runs and changes pageableQueryParam. This causes (1) to re-run, aborting the previous run of (1). During this abort, showProcessing set to true in run and re-run of (1) is prematurely set to false in the finally block. This shows the empty table flash.


amritkv (approved):
LGTM !

PR #2069, opened 2026-09-17
fix(Project): Add server-side sorting to SearchReleasesModal component

Farooq-Fateh-Aftab:
The `Link Releases` modal table had no sorting controls in its column headers, unlike the `Link Projects` modal.
This PR adds server-side sorting to the `Link Releases` table, implemented exactly the same way as in `LinkProjectsModal`.

Changes in `src/components/sw360/SearchReleasesModal.tsx`:

* Added `accessorKey` / `accessorFn` to the data columns (`vendor`, `name`, `version`, `clearingState`, `mainlineState`),
  which is what makes `SW360Table` render the sort indicator in the header. The select checkbox column has no accessor
  and therefore stays unsortable.
* Renamed column ids to match the backend sort fields (`componentName` -> `name`, `releaseVersion` -> `version`).
* Added `manualSorting: true` and an `onSortingChange` handler that stores the sort as `"<field>,<asc|desc>"` in
  `pageableQueryParam`, which triggers the already existing server-side search request.
* Derived `state.sorting` from `pageableQueryParam.sort` so the header icon reflects the active sort.

No new dependencies were added.

Issue: https://github.com/eclipse-sw360/sw360-frontend/issues/2068

### How To Test?
1. Open a project in edit mode and go to the `Linked Releases and Projects` tab.
2. Click `Add Releases` to open the `Link Releases` modal.
3. Enter a search term and press `Search`.
4. Verify that every column header except the checkbox column shows a sort icon.
5. Click a column header and verify:
   * The results are re-fetched and ordered by that column ascending.
   * Clicking again switches to descending, and the icon changes accordingly.
   * Sorting is preserved while paging through the result set and when changing the page size.
6. Verify that selecting releases and `Link Releases` still works after sorting.


amritkv (approved):
LGTM !

Issue #2070, opened 2026-09-17
When deleting one package-url, other package-urls are deleted!


Shivani2118:
When I delete a package-url, other package-urls get deleted as well.

**How to reproduce:**
1. Go to the release version 
2. Click "Edit"
3. Delete one package-url.
4. Click "Update release"
5. All package-urls were deleted except one. The other package-urls can only be saved via the data from the Changelog.

amritkv:
Closing this issue as the PR https://github.com/eclipse-sw360/sw360-frontend/pull/2071 is merged.

PR #2071, opened 2026-09-17
fix: Prevent deleting one package-url from removing all others

Shivani2118:
#2070 
Deleting one package-url from a release was wiping out all the other package-urls, leaving only one entry after saving.


src/utils/common.utils.ts
```diff
@@ -149,6 +149,34 @@ const convertObjectToMap = (data: { [k: string]: string }): InputKeyValue[] => {
...
+/**
+ * Converts an array of key-value pairs to a map.
+ * Special handling for 'package-url' key: multiple entries with this key are
+ * combined into a single JSON array string instead of overwriting one another.
+ * @param list - The array of key-value pairs to convert.
+ * @returns A map of key-value pairs.
+ */
+const convertInputListToMap = (list: InputKeyValue[]): Map<string, string> => {
+    const map = new Map<string, string>()
+    const packageUrls: string[] = []
```
amritkv: Hey @Shivani2118 !
As this is a common component, so it would be good to rename the var `packageUrls` to some generic one.

src/components/sw360/AddKeyValue/AddKeyValue.tsx
```diff
@@ -49,26 +49,7 @@ function AddKeyValue(props: Props): JSX.Element {
...
-            }
-        })
-
-        // Combine package-url values into a JSON array string
-        if (packageUrls.length > 0) {
-            map.set('package-url', JSON.stringify(packageUrls))
-        }
-
-        return map
+        return CommonUtils.convertInputListToMap(list)
```
amritkv: As this is single liner function, we can directly call `CommonUtils.convertInputListToMap(list)` from the calling function. No need to have a separate function for that.

amritkv (changes requested):

amritkv (approved):
LGTM !

PR #2081, opened 2026-09-20
fix(docker): Enforce LF line endings for config and secret files

Neet2516:
### Problem
When cloning this repository on Windows (where Git defaults to `core.autocrlf = true`), files in `config/` (such as `config/couchdb/default_secrets`) are checked out with CRLF (`\r\n`) line endings.

When these files are sourced inside the Linux Docker container (e.g., `source /run/secrets/COUCHDB_SECRETS`), the carriage return (`\r`) gets appended to environment variables (e.g., `COUCHDB_PASSWORD="sw360fossie\r"`).

### Impact
- Backend fails to authenticate with CouchDB due to the trailing carriage return.
- CouchDB locks the account (`Account is temporarily locked due to multiple authentication failures`).
- Required databases (`sw360users`, etc.) fail to initialize, leaving the backend container unhealthy and causing frontend login failures (`Error while fetching User Group`).

### Solution
- Added `.gitattributes` to enforce Unix LF line endings (`eol=lf`) on shell scripts, configs, and secret files regardless of host OS Git settings.


GMishx (approved):
Good to have

PR #2082, opened 2026-09-22
Improve performance of Generate License Info Page

deo002:


deo002:
Closing in favour of #2087 .

With react-virtual, we could not do "Ctrl + F" on components not present on screen. Also, the performance gain from react-virtual and logic handover to tanstack was still not on par with what we got from shifting to uncontrolled components.

PR #2085, opened 2026-09-23
fix(clear): Use button icon instead of x

GMishx:
Instead of following
image

Normalize everything to following
image


amritkv (approved):
LGTM !

PR #2086, opened 2026-09-23
fix(css): Fix cursor-pointer CSS

GMishx:
1. Make the `cursor: pointer` class more specific to override form-controller cursor change.
2. Add this class to other inputs where it should be used. For example Roles where you need to click on input element to open search pop-up.


amritkv (approved):
LGTM !

